PR #17, opened 2023-06-29
redirect validation resource

wesleymccollam:


henryrecker-pingidentity:
Do we have the HCL formatting linter in this repo? The spacing of the HCL example looks a little off

internal/provider/provider.go
```diff
@@ -23,6 +23,7 @@ import (
 	oauthAuthServerSettingsScopesExclusiveScopes "github.com/pingidentity/terraform-provider-pingfederate/internal/resource/oauthauthserversettingsscopesexclusivescopes"
 	oauthIssuer "github.com/pingidentity/terraform-provider-pingfederate/internal/resource/oauthissuers"
 	protocolMetadataLifetimeSettings "github.com/pingidentity/terraform-provider-pingfederate/internal/resource/protocolmetadatalifetimesettings"
+	redirectValidation "github.com/pingidentity/terraform-provider-pingfederate/internal/resource/redirectvalidation"
```
henryrecker-pingidentity, line 26: We probably don't need to rename these imports

internal/types/client_utils.go
```diff
@@ -37,6 +37,74 @@ func ToStateAuthenticationPolicyContract(r *client.AuthenticationPolicyContract)
 	return caSliceOfObj, eaSliceOfObj
 }
 
+func AttrValueToBoolPointer(val attr.Value) *bool {
+	boolPointerVal := ConvertToPrimitive(val).(bool)
```
henryrecker-pingidentity, line 41: Probably better to do this by converting to a types.Bool and then using the ValueBoolPointer method

internal/resource/redirectvalidation/redirect_validation_resource.go
```diff
@@ -0,0 +1,417 @@
...
+	r.providerConfig = providerCfg.ProviderConfig
+	r.apiClient = providerCfg.ApiClient
+
+}
+
+func readRedirectValidationResponse(ctx context.Context, r *client.RedirectValidationSettings, state *redirectValidationResourceModel, expectedValues *redirectValidationResourceModel) {
+	state.Id = types.StringValue("id")
+	stateRedirectValidationLocalSettings, _ := internaltypes.ToStateRedirectValidation(r, diag.Diagnostics{})
+	state.RedirectValidationLocalSettings = stateRedirectValidationLocalSettings
+	_, stateRedirectValidationPartnerSettings := internaltypes.ToStateRedirectValidation(r, diag.Diagnostics{})
```
henryrecker-pingidentity, line 273: You could combine 271 and 273 into one call so you don't run the method twice

internal/resource/redirectvalidation/redirect_validation_resource.go
```diff
@@ -0,0 +1,417 @@
...
+		return
+	}
+
+	providerCfg := req.ProviderData.(internaltypes.ResourceConfiguration)
+	r.providerConfig = providerCfg.ProviderConfig
+	r.apiClient = providerCfg.ApiClient
+
+}
+
+func readRedirectValidationResponse(ctx context.Context, r *client.RedirectValidationSettings, state *redirectValidationResourceModel, expectedValues *redirectValidationResourceModel) {
```
henryrecker-pingidentity, line 269: Looks like expectedValues isn't needed in this method

internal/types/client_utils.go
```diff
@@ -37,6 +37,74 @@ func ToStateAuthenticationPolicyContract(r *client.AuthenticationPolicyContract)
...
+	return &boolPointerVal
+}
+
+func ToStateRedirectValidation(r *client.RedirectValidationSettings, diags diag.Diagnostics) (basetypes.ObjectValue, basetypes.ObjectValue) {
```
henryrecker-pingidentity, line 45: If this only applies to this new resource I would move it to the resource Go file

internal/resource/redirectvalidation/redirect_validation_resource.go
wesleymccollam: Fixed in the latest commit

internal/types/client_utils.go
wesleymccollam: Fixed in the latest commit

internal/resource/redirectvalidation/redirect_validation_resource.go
wesleymccollam: This was moved to the resource file in the latest commit.

internal/types/client_utils.go
wesleymccollam: Removed this completely as it's not used.

internal/provider/provider.go
wesleymccollam: Imports were fixed as a part of another branch recently merged into main. These have been merged and the new structure is applied in the latest commit for this resource.

internal/acctest/acctest.go
```diff
@@ -153,6 +153,13 @@ func TestAttributesMatchBool(resourceType string, resourceName *string, attribut
 	return nil
 }
 
+func TestAttributesMatchBoolPointer(resourceType string, resourceName *string, attributeName string, expected, found *bool) error {
+	if expected != found {
+		return mismatchedAttributeError(resourceType, resourceName, attributeName, strconv.FormatBool(*expected), strconv.FormatBool(*found))
```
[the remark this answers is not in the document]
wesleymccollam: Removed this completely as it's not used.

henryrecker-pingidentity (approved):

PR #30, opened 2023-08-02
System Keys resource

jith-george:


internal/resource/config/serversettings/server_settings_system_keys_resource.go
```diff
@@ -0,0 +1,396 @@
...
+func serverSettingsSystemKeysResourceSchema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse, setOptionalToComputed bool) {
+	schema := schema.Schema{
+		Description: "Manages a Server Settings SystemKeys.",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Description: "Placeholder for Terraform",
+				Computed:    true,
+				Optional:    false,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
```
henryrecker-pingidentity, line 62: We probably don't need this here

internal/resource/config/serversettings/server_settings_system_keys_resource.go
```diff
@@ -0,0 +1,396 @@
...
+						PlanModifiers: []planmodifier.String{
+							stringplanmodifier.UseStateForUnknown(),
+						},
+					},
+					"key_data": schema.StringAttribute{
+						Description: "The clear text system key base 64 encoded. The system key must be 32 bytes before base 64 encoding",
+						Computed:    true,
+						Optional:    false,
+						PlanModifiers: []planmodifier.String{
+							stringplanmodifier.UseStateForUnknown(),
```
henryrecker-pingidentity, line 91: If this is computed fully by PF and we don't need to pass it to the API, then we shouldn't need UseStateForUnknown, same for the other key_data fields

internal/resource/config/serversettings/server_settings_system_keys_resource.go
```diff
@@ -0,0 +1,396 @@
...
+				Computed:    true,
+				Optional:    false,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"current": schema.SingleNestedAttribute{
+				Description: "Current SystemKeys Secrets that are used in cryptographic operations to generate and consume internal tokens.",
+				Required:    true,
+				PlanModifiers: []planmodifier.Object{
```
henryrecker-pingidentity, line 68: If this is required then we don't need UseStateForUnknown, same for pending

internal/resource/config/serversettings/server_settings_system_keys_resource.go
```diff
@@ -0,0 +1,396 @@
...
+						PlanModifiers: []planmodifier.String{
+							stringplanmodifier.UseStateForUnknown(),
+						},
+					},
+				},
+			},
+		},
+	}
+
+	resp.Schema = schema
```
henryrecker-pingidentity, line 162: If you don't need to make any changes to schema down here you can just make this the top line of the method. `resp.Schema = schema.Schema{...`. Rather than saving it to a separate variable like this.

internal/resource/config/serversettings/server_settings_system_keys_resource.go
```diff
@@ -0,0 +1,396 @@
...
+}
+
+func addServerSettingsSystemKeysFields(ctx context.Context, addRequest *client.SystemKeys, plan serverSettingsSystemKeysResourceModel) error {
+
+	if internaltypes.IsDefined(plan.Current) {
+		currentAttrs := plan.Current.Attributes()
+		encryptedKeyDataAttrcurrent := currentAttrs["encrypted_key_data"]
+		if internaltypes.IsNonEmptyString(encryptedKeyDataAttrcurrent.(types.String)) {
+			addRequest.Current = *client.NewSystemKey()
+			currentEncryptedKeyData := internaltypes.ConvertToPrimitive(encryptedKeyDataAttrcurrent).(string)
```
henryrecker-pingidentity, line 172: Since we know encryptedKeyDataAttrPrevious is a types.String here we don't need this extra casting really. You can cast it to a types.String and use the ValueString() method. I think that would be clearer here. same for the below in this method

internal/resource/config/serversettings/server_settings_system_keys_resource.go
```diff
@@ -0,0 +1,396 @@
...
+	if internaltypes.IsDefined(plan.Pending) {
+		pendingAttrs := plan.Pending.Attributes()
+		encryptedKeyDataAttrPending := pendingAttrs["encrypted_key_data"]
+		if internaltypes.IsNonEmptyString(encryptedKeyDataAttrPending.(types.String)) {
+			addRequest.Pending = *client.NewSystemKey()
+			pendingEncryptedKeyData := internaltypes.ConvertToPrimitive(encryptedKeyDataAttrPending).(string)
+			addRequest.Pending.EncryptedKeyData = &pendingEncryptedKeyData
+		}
+	}
+	return nil
```
henryrecker-pingidentity, line 194: If the method always returns nil then there's no reason to have the method return anything at all

internal/resource/config/serversettings/server_settings_system_keys_resource.go
```diff
@@ -0,0 +1,396 @@
...
+// This config object is edit-only, so Terraform can't delete it.
+func (r *serverSettingsSystemKeysResource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
+}
+
+func (r *serverSettingsSystemKeysResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	importServerSettingsSystemKeysLocation(ctx, req, resp)
+}
+func importServerSettingsSystemKeysLocation(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	// Retrieve import ID and save to id attribute
+	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
```
henryrecker-pingidentity, line 395: Might be good to put a comment here saying that the id provided doesn't matter, since this appears to be a singleton resource

henryrecker-pingidentity (changes requested):

henryrecker-pingidentity (approved):

PR #34, opened 2023-08-21
oauth auth server settings resource

wesleymccollam:
- oauth auth server settings resource
- internal methods
- descriptions added to administrative accounts
- Makefile acceptance test parallelism set from 4 to 1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ testacc:
 	PINGFEDERATE_PROVIDER_USERNAME=administrator \
 	PINGFEDERATE_PROVIDER_PASSWORD=2FederateM0re \
 	PINGFEDERATE_PROVIDER_INSECURE_TRUST_ALL_TLS=true \
-	TF_ACC=1 go test -timeout 20m -v ./internal/... -p 1
+	TF_ACC=1 go test -timeout 10m -v ./internal/... -p 1
 
 testacccomplete: spincontainer testacc
 
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ testacc:
 	PINGFEDERATE_PROVIDER_USERNAME=administrator \
 	PINGFEDERATE_PROVIDER_PASSWORD=2FederateM0re \
 	PINGFEDERATE_PROVIDER_INSECURE_TRUST_ALL_TLS=true \
-	TF_ACC=1 go test -timeout 10m -v ./internal/... -p 4
+	TF_ACC=1 go test -timeout 20m -v ./internal/... -p 1
 
 testacccomplete: spincontainer testacc
 
@@ -69,8 +69,11 @@ generateresource:
 	PINGFEDERATE_GENERATED_ENDPOINT=oauth/authServerSettings/scopes/exclusiveScopes \
 	PINGFEDERATE_RESOURCE_DEFINITION_NAME=ScopeEntry \
 	PINGFEDERATE_ALLOW_REQUIRED_BYPASS=False \
+	OVERWRITE_EXISTING_RESOURCE_FILE=False \
 	PINGFEDERATE_PUT_ONLY_RESOURCE=False \
+	GENERATE_SCHEMA=True \
 	python3 scripts/generate_resource.py
+	make fmt
 	
 openlocalwebapi:
 	open "https://localhost:9999/pf-admin-api/api-docs/#/"
--- a/internal/resource/config/oauth/oauth_auth_server_settings_resource.go
+++ b/internal/resource/config/oauth/oauth_auth_server_settings_resource.go
@@ -1,0 +1,1167 @@
+package oauth
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/boolplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/int64planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/objectplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/setplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	client "github.com/pingidentity/pingfederate-go-client"
+	internaljson "github.com/pingidentity/terraform-provider-pingfederate/internal/json"
+	"github.com/pingidentity/terraform-provider-pingfederate/internal/resource/config"
+	internaltypes "github.com/pingidentity/terraform-provider-pingfederate/internal/types"
+)
+
+// Ensure the implementation satisfies the expected interfaces.
+var (
+	_ resource.Resource                = &oauthAuthServerSettingsResource{}
+	_ resource.ResourceWithConfigure   = &oauthAuthServerSettingsResource{}
+	_ resource.ResourceWithImportState = &oauthAuthServerSettingsResource{}
+)
+
+// OauthAuthServerSettingsResource is a helper function to simplify the provider implementation.
+func OauthAuthServerSettingsResource() resource.Resource {
+	return &oauthAuthServerSettingsResource{}
+}
+
+// oauthAuthServerSettingsResource is the resource implementation.
+type oauthAuthServerSettingsResource struct {
+	providerConfig internaltypes.ProviderConfiguration
+	apiClient      *client.APIClient
+}
+
+type oauthAuthServerSettingsResourceModel struct {
+	Id                                          types.String `tfsdk:"id"`
+	DefaultScopeDescription                     types.String `tfsdk:"default_scope_description"`
+	Scopes                                      types.Set    `tfsdk:"scopes"`
+	ScopeGroups                                 types.Set    `tfsdk:"scope_groups"`
+	ExclusiveScopes                             types.Set    `tfsdk:"exclusive_scopes"`
+	ExclusiveScopeGroups                        types.Set    `tfsdk:"exclusive_scope_groups"`
+	AuthorizationCodeTimeout                    types.Int64  `tfsdk:"authorization_code_timeout"`
+	AuthorizationCodeEntropy                    types.Int64  `tfsdk:"authorization_code_entropy"`
+	DisallowPlainPKCE                           types.Bool   `tfsdk:"disallow_plain_pkce"`
+	IncludeIssuerInAuthorizationResponse        types.Bool   `tfsdk:"include_issuer_in_authorization_response"`
+	TrackUserSessionsForLogout                  types.Bool   `tfsdk:"track_user_sessions_for_logout"`
+	TokenEndpointBaseUrl                        types.String `tfsdk:"token_endpoint_base_url"`
+	PersistentGrantLifetime                     types.Int64  `tfsdk:"persistent_grant_lifetime"`
+	PersistentGrantLifetimeUnit                 types.String `tfsdk:"persistent_grant_lifetime_unit"`
+	PersistentGrantIdleTimeout                  types.Int64  `tfsdk:"persistent_grant_idle_timeout"`
+	PersistentGrantIdleTimeoutTimeUnit          types.String `tfsdk:"persistent_grant_idle_timeout_time_unit"`
+	RefreshTokenLength                          types.Int64  `tfsdk:"refresh_token_length"`
+	RollRefreshTokenValues                      types.Bool   `tfsdk:"roll_refresh_token_values"`
+	RefreshTokenRollingGracePeriod              types.Int64  `tfsdk:"refresh_token_rolling_grace_period"`
+	RefreshRollingInterval                      types.Int64  `tfsdk:"refresh_rolling_interval"`
+	PersistentGrantReuseGrantTypes              types.Set    `tfsdk:"persistent_grant_reuse_grant_types"`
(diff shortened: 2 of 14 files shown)


internal/resource/config/oauth/oauth_auth_server_settings_resource.go
```diff
@@ -0,0 +1,1167 @@
...
+			"client_secret_retention_period": schema.Int64Attribute{
+				Description: "The length of time in minutes that client secrets will be retained as secondary secrets after secret change. The default value is 0, which will disable secondary client secret retention.",
+				Computed:    true,
+				Optional:    true,
+				PlanModifiers: []planmodifier.Int64{
+					int64planmodifier.UseStateForUnknown(),
+				},
+			},
+			"jwt_secured_authorization_response_mode_lifetime": schema.Int64Attribute{
+				Description: "The lifetime, in seconds, of the JWT Secured authorization response. The default value is 600.",
```
henryrecker-pingidentity, line 572: It looks like we know the defaults for a lot of these attributes, so it would be nice to set them in the schema here to have more consistent provider runs across versions.

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
```diff
@@ -0,0 +1,1167 @@
...
+	ApprovedAuthorizationDetailAttribute        types.String `tfsdk:"approved_authorization_detail_attribute"`
+	ParReferenceTimeout                         types.Int64  `tfsdk:"par_reference_timeout"`
+	ParReferenceLength                          types.Int64  `tfsdk:"par_reference_length"`
+	ParStatus                                   types.String `tfsdk:"par_status"`
+	ClientSecretRetentionPeriod                 types.Int64  `tfsdk:"client_secret_retention_period"`
+	JwtSecuredAuthorizationResponseModeLifetime types.Int64  `tfsdk:"jwt_secured_authorization_response_mode_lifetime"`
+}
+
+// GetSchema defines the schema for the resource.
+func (r *oauthAuthServerSettingsResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
```
henryrecker-pingidentity, line 98: We don't need to split this into separate methods like the PD provider does if we don't have default_ versions of resources. The same applies to the other methods (read, update, etc.)

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
```diff
@@ -0,0 +1,1167 @@
...
+	if internaltypes.IsDefined(model.Scopes) {
+		scopeElems := model.Scopes.Elements()
+		for _, scopeElem := range scopeElems {
+			scopeElemObjectAttrs := scopeElem.(types.Object)
+			scopeEntryName := scopeElemObjectAttrs.Attributes()["name"].(basetypes.StringValue).ValueString()
+			scopeNames = append(scopeNames, scopeEntryName)
+			scopeEntryIsDynamic := scopeElemObjectAttrs.Attributes()["dynamic"].(basetypes.BoolValue).ValueBool()
+			if scopeEntryIsDynamic {
+				if string(scopeEntryName[0]) != "*" {
+					resp.Diagnostics.AddError("Scope name conflict!", fmt.Sprintf("Scope name \"%s\" must be prefixed, suffixed or both by static part(s) with one and only one dynamic part represented by \"*\"", scopeEntryName))
```
henryrecker-pingidentity, line 620: This error message is confusing to me compared to what the code is checking, which is just that the first character is an asterisk.

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
```diff
@@ -0,0 +1,1167 @@
...
+	// Test scope names for dynamic true, string must be prepended with *
+	if internaltypes.IsDefined(model.Scopes) {
+		scopeElems := model.Scopes.Elements()
+		for _, scopeElem := range scopeElems {
+			scopeElemObjectAttrs := scopeElem.(types.Object)
+			scopeEntryName := scopeElemObjectAttrs.Attributes()["name"].(basetypes.StringValue).ValueString()
+			scopeNames = append(scopeNames, scopeEntryName)
+			scopeEntryIsDynamic := scopeElemObjectAttrs.Attributes()["dynamic"].(basetypes.BoolValue).ValueBool()
+			if scopeEntryIsDynamic {
+				if string(scopeEntryName[0]) != "*" {
```
henryrecker-pingidentity, line 619: Add a check for string length >= 1 here to avoid a panic.

Also I'm not sure if this is a safe way to get the first character (rune) of a string in go - it might not handle non-ASCII characters well. We could write some utility function that converts a string to a slice of runes potentially.

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
```diff
@@ -0,0 +1,1167 @@
...
+	if internaltypes.IsDefined(model.ExclusiveScopes) {
+		exclusiveScopeElems := model.ExclusiveScopes.Elements()
+		for _, esElem := range exclusiveScopeElems {
+			esElemObjectAttrs := esElem.(types.Object)
+			eScopeEntryName := strings.ToLower(esElemObjectAttrs.Attributes()["name"].(basetypes.StringValue).ValueString())
+			eScopeNames = append(eScopeNames, eScopeEntryName)
+			eScopeEntryIsDynamic := esElemObjectAttrs.Attributes()["dynamic"].(basetypes.BoolValue).ValueBool()
+			if eScopeEntryIsDynamic {
+				if string(eScopeEntryName[0]) != "*" {
+					resp.Diagnostics.AddError("Dynamic scope name conflict!", fmt.Sprintf("Dynamic scope name \"%s\" must be prefixed, suffixed or both by static part(s) with one and only one dynamic part represented by \"*\"", eScopeEntryName))
```
henryrecker-pingidentity, line 637: Same comments as the above

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
```diff
@@ -0,0 +1,1167 @@
...
+
+	if internaltypes.IsDefined(plan.Scopes) {
+		err := json.Unmarshal([]byte(internaljson.FromValue(plan.Scopes)), &addRequest.Scopes)
+		if err != nil {
+			return err
+		}
+	}
+
+	if internaltypes.IsDefined(plan.ScopeGroups) {
+		addRequest.ScopeGroups = client.NewAuthorizationServerSettingsWithDefaults().ScopeGroups
```
henryrecker-pingidentity, line 664: Why is this needed?

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
```diff
@@ -0,0 +1,1167 @@
...
+	}
+
+	providerCfg := req.ProviderData.(internaltypes.ResourceConfiguration)
+	r.providerConfig = providerCfg.ProviderConfig
+	r.apiClient = providerCfg.ApiClient
+
+}
+
+func readOauthAuthServerSettingsResponse(ctx context.Context, r *client.AuthorizationServerSettings, state *oauthAuthServerSettingsResourceModel) {
+	state.Id = types.StringValue("id")
```
henryrecker-pingidentity, line 870: Is there a unique value in the response that would be usable here? Ideally we can have different ids for different resources of the same type.

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
```diff
@@ -0,0 +1,1167 @@
...
+	}
+	state.ExclusiveScopeGroups, _ = types.SetValueFrom(ctx, types.ObjectType{AttrTypes: exclusiveScopeGroupAttrTypes}, toStateExclusiveScopeGroups)
+
+	// state.PersistentGrantContract
+	getPersistentGrantContract := r.GetPersistentGrantContract()
+	// Build core attributes
+	nameAttributeType := map[string]attr.Type{
+		"name": basetypes.StringType{},
+	}
+	coreAttrs := getPersistentGrantContract.CoreAttributes
```
henryrecker-pingidentity, line 951: Is there any chance this could be nil?

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
```diff
@@ -0,0 +1,1167 @@
...
+	// Test if values in sets match
+	matchNameBtwnScopes, matchVal := internaltypes.MatchStringInSets(scopeNames, eScopeNames)
+	if matchNameBtwnScopes {
+		resp.Diagnostics.AddError("Scope name conflict!", fmt.Sprintf("The scope name \"%s\" is already defined in another scope list", matchVal))
+	}
+}
+
+func addOptionalOauthAuthServerSettingsFields(ctx context.Context, addRequest *client.AuthorizationServerSettings, plan oauthAuthServerSettingsResourceModel) error {
+
+	if internaltypes.IsDefined(plan.DefaultScopeDescription) {
```
henryrecker-pingidentity, line 652: It looks like some of the attributes in this method such as this one aren't optional, so we don't need to re-add them to the addrequest here.

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
```diff
@@ -0,0 +1,1167 @@
...
+
+	diags := req.State.Get(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	apiReadOauthAuthServerSettings, httpResp, err := apiClient.OauthAuthServerSettingsApi.GetAuthorizationServerSettings(config.ProviderBasicAuthContext(ctx, providerConfig)).Execute()
+
+	if err != nil {
+		config.ReportHttpError(ctx, &resp.Diagnostics, "An error occurred while looking for a OauthAuthServerSettings", err, httpResp)
```
henryrecker-pingidentity, line 1082: As I learned recently, we shouldn't report an error on read for a 404 https://developer.hashicorp.com/terraform/plugin/framework/resources/read#recommendations

The same applies to delete  https://developer.hashicorp.com/terraform/plugin/framework/resources/delete#recommendations

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
```diff
@@ -0,0 +1,1167 @@
...
+	if err == nil {
+		tflog.Debug(ctx, "Read response: "+string(responseJson))
+	}
+	// Read the response
+	readOauthAuthServerSettingsResponse(ctx, updateOauthAuthServerSettingsResponse, &state)
+
+	// Update computed values
+	diags = resp.State.Set(ctx, state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
```
henryrecker-pingidentity, line 1148: I realized these blocks were pointless recently so I removed them from the PD resources (as in the if statement -> return blocks at the very end of methods where they don't actually skip any logic)

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
```diff
@@ -0,0 +1,1167 @@
...
+				Optional:    true,
+				PlanModifiers: []planmodifier.Int64{
+					int64planmodifier.UseStateForUnknown(),
+				},
+			},
+		},
+	}
+
+	// Set attributes in string list
+	if setOptionalToComputed {
```
henryrecker-pingidentity, line 583: Looks like this is never going to run because the boolean is always false

henryrecker-pingidentity (changes requested):

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
henryrecker-pingidentity: Nevermind I see this is an edit-only resource based on the comment on the delete method.

Makefile
```diff
@@ -52,7 +52,7 @@ testacc:
 	PINGFEDERATE_PROVIDER_USERNAME=administrator \
 	PINGFEDERATE_PROVIDER_PASSWORD=2FederateM0re \
 	PINGFEDERATE_PROVIDER_INSECURE_TRUST_ALL_TLS=true \
-	TF_ACC=1 go test -timeout 10m -v ./internal/... -p 4
+	TF_ACC=1 go test -timeout 20m -v ./internal/... -p 1
```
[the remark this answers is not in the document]
wesleymccollam: Just wanted to match the PD provider

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
wesleymccollam: these are the values that are always returned from the client

Makefile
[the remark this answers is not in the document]
wesleymccollam: Parallelism fixes tests that seem to conflict

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
henryrecker-pingidentity: This one is fixed but there are similar ones below that start with defaults rather than reading straight from the plan

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
```diff
@@ -0,0 +1,1167 @@
...
+	schema := schema.Schema{
+		Description: "Manages Oauth Auth Server Settings",
+		Attributes: map[string]schema.Attribute{
+			"default_scope_description": schema.StringAttribute{
+				Description: "The default scope description.",
+				Required:    true,
+			},
+			"scopes": schema.SetNestedAttribute{
+				Description: "The list of common scopes.",
+				Computed:    true,
```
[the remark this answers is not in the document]
wesleymccollam: In the latest commit, all computed values were removed from the schema, tf apply, then added back as needed.

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
wesleymccollam: It wasn't 🥇

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
wesleymccollam: Add unnecessary requests were removed in the latest commit.

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
wesleymccollam: 404 check added in the latest commit

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
wesleymccollam: now removed where applicable

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
wesleymccollam: doh! 👍 , removed

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
wesleymccollam: the remaining properties set addRequest to use the type with defaults as properties within the property object are not nullable

henryrecker-pingidentity (approved):

PR #38, opened 2023-08-31
Local Identity Profile Resource

jith-george:


internal/resource/config/localidentity/local_identity_identity_profiles_resource.go
```diff
@@ -0,0 +1,917 @@
...
+	// Validates Email Verification type for Email Configuration
+	if internaltypes.IsDefined(model.EmailVerificationConfig) {
+		emailVerificationConfig := model.EmailVerificationConfig.Attributes()
+		emailVerificationType := model.EmailVerificationConfig.Attributes()["email_verification_type"].(basetypes.StringValue).ValueString()
+		switch emailVerificationType {
+		case "OTP":
+			if internaltypes.IsDefined(emailVerificationConfig["otl_time_to_live"].(basetypes.Int64Value)) {
+				resp.Diagnostics.AddError("Invalid Attribute Combination!", fmt.Sprintln("otl_time_to_live attribute is not allowed when email_verification_type is OTP. Required attributes are otp_length, otp_retry_attempts and otp_time_to_live."))
+			}
+			if internaltypes.IsDefined(emailVerificationConfig["otp_length"].(basetypes.Int64Value)) && (emailVerificationConfig["otp_length"].(basetypes.Int64Value).ValueInt64() < 5) {
```
henryrecker-pingidentity, line 576: If someone was to pass in a non-int value for otp_length in HCL for example, would this cause a panic? I don't think it would but it would be good to check. It feels a little risky to have all these casts everywhere.

internal/resource/config/localidentity/local_identity_identity_profiles_resource.go
```diff
@@ -0,0 +1,917 @@
...
+	r.apiClient = providerCfg.ApiClient
+
+}
+func (r *localIdentityIdentityProfilesResource) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
+
+	var model localIdentityIdentityProfilesResourceModel
+	resp.Diagnostics.Append(req.Config.Get(ctx, &model)...)
+	// Validates Email Verification type for Email Configuration
+	if internaltypes.IsDefined(model.EmailVerificationConfig) {
+		emailVerificationConfig := model.EmailVerificationConfig.Attributes()
```
henryrecker-pingidentity, line 569: Rather than casting everything it might be easier and more readable if you convert this into the client struct type, so you can access the fields directly.

The same applies below when checking other objects' fields

internal/resource/config/localidentity/local_identity_identity_profiles_resource.go
```diff
@@ -0,0 +1,917 @@
...
+				resp.Diagnostics.AddError("Invalid Attribute Combination!", fmt.Sprintln("otp_length attribute is not allowed when email_verification_type is OTL. Required attribute: otl_time_to_live."))
+			}
+			if internaltypes.IsDefined(emailVerificationConfig["email_verification_otp_template_name"].(basetypes.StringValue)) {
+				resp.Diagnostics.AddError("Invalid Attribute Combination!", fmt.Sprintln("email_verification_otp_template_name attribute is not allowed when email_verification_type is OTL. Required attribute: otl_time_to_live."))
+			}
+		}
+	}
+
+	if (internaltypes.IsDefined(model.ProfileEnabled) && (!model.ProfileEnabled.ValueBool())) && (internaltypes.IsDefined(model.RegistrationEnabled) && (!model.RegistrationEnabled.ValueBool())) {
+		resp.Diagnostics.AddError("Invalid Attribute Combination!", fmt.Sprintln("email, data_store_config, fields, registration_config and profile_config are not allowed when registration and profile are disabled."))
```
henryrecker-pingidentity, line 605: It doesn't look to me like this if statement is checking what the error says. The if checks if profile_enabled is false and registration_enabled is false, but it doesn't check any of those other fields mentioned.

internal/resource/config/localidentity/local_identity_identity_profiles_resource.go
```diff
@@ -0,0 +1,917 @@
...
+			if (model.RegistrationEnabled.ValueBool()) && (model.RegistrationConfig.IsNull()) {
+				resp.Diagnostics.AddError("Invalid Value for Attribute!", fmt.Sprintln("registration_config is required when registration is enabled."))
+			}
+		}
+	}
+}
+
+func readLocalIdentityIdentityProfilesResponse(ctx context.Context, r *client.LocalIdentityProfile, state *localIdentityIdentityProfilesResourceModel) {
+	state.Id = internaltypes.StringTypeOrNil(r.Id, false)
+	state.Name = types.StringPointerValue(&r.Name)
```
henryrecker-pingidentity, line 658: You're converting to a pointer and then back to a string here. You can just use `types.StringValue(r.Name)`

internal/resource/config/localidentity/local_identity_identity_profiles_resource.go
```diff
@@ -0,0 +1,917 @@
...
+	httpResp, err := r.apiClient.LocalIdentityIdentityProfilesApi.DeleteIdentityProfile(config.ProviderBasicAuthContext(ctx, r.providerConfig), state.Id.ValueString()).Execute()
+	if err != nil {
+		config.ReportHttpError(ctx, &resp.Diagnostics, "An error occurred while deleting Local Identity Profile", err, httpResp)
+		return
+	}
+
+}
+func (r *localIdentityIdentityProfilesResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	// Retrieve import ID and save to id attribute
+	// Set a placeholder id value to appease terraform.
```
henryrecker-pingidentity, line 913: This comment is off, in this case we aren't using a placeholder id.

internal/resource/config/localidentity/local_identity_identity_profiles_resource.go
```diff
@@ -0,0 +1,917 @@
...
+	// Retrieve values from plan
+	var plan localIdentityIdentityProfilesResourceModel
+	diags := req.Plan.Get(ctx, &plan)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	// Get the current state to see how any attributes are changing
+	var state localIdentityIdentityProfilesResourceModel
+	req.State.Get(ctx, &state)
```
henryrecker-pingidentity, line 862: I'm not sure you need this line. Depending on if the PF API returns everything in the response every time.

henryrecker-pingidentity (changes requested):

examples/resources/pingfederate_local_identity_identity_profiles/resource.tf
```diff
@@ -0,0 +1,134 @@
...
+          "Mask Log Values" = false,
+        }
+      },
+    ]
+    strip_space_from_unique_field = true
+  }
+  email_verification_config = {
+    email_verification_enabled = true
+    verify_email_template_name = "message-template-email-ownership-verification.html"
+    /* email_verification_sent_template_name = "local.identity.email.verification.sent.html"  */
```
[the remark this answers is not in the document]
jith-george: They are supposed to be there if someone changes the email_verification_type from "OTP" to "OTL"

internal/resource/config/localidentity/local_identity_identity_profiles_resource.go
```diff
@@ -0,0 +1,917 @@
...
+func (r *localIdentityIdentityProfilesResource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
+	// Retrieve values from state
+	var state localIdentityIdentityProfilesResourceModel
+	diags := req.State.Get(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	httpResp, err := r.apiClient.LocalIdentityIdentityProfilesApi.DeleteIdentityProfile(config.ProviderBasicAuthContext(ctx, r.providerConfig), state.Id.ValueString()).Execute()
+	if err != nil {
```
henryrecker-pingidentity, line 905: This should ignore errors that signify the resource doesn't exist: https://developer.hashicorp.com/terraform/plugin/framework/resources/delete#recommendations
Also the return isn't needed here

internal/acctest/localidentityidentityprofiles/local_identity_identity_profiles_test.go
```diff
@@ -0,0 +1,141 @@
...
+		id:                  localIdentityIdentityProfilesId,
+		name:                "example",
+		registrationEnabled: false,
+		profileEnabled:      false,
+	}
+	updatedResourceModel := localIdentityIdentityProfilesResourceModel{
+		id:                  localIdentityIdentityProfilesId,
+		name:                "example1",
+		registrationEnabled: false,
+		profileEnabled:      false,
```
henryrecker-pingidentity, line 38: We still aren't covering any nested attributes here, which I think are important to test, like some of the config attributes. Also it's best if we change the values between the initialResourceModel and updatedResourceModel if possible.

jith-george: This resource would need a Directory dataStore to test the nested attributes.
Separate ticket opened to improve the test for nested attributes once we complete the DataStore resource.

henryrecker-pingidentity (approved):

PR #53, opened 2023-09-22
code and test improvements

wesleymccollam:
- remove panic's
- test parallelism increase
- gosec pipeline result output


internal/resource/config/oauth/oauth_auth_server_settings_resource.go
```diff
@@ -53,7 +54,7 @@ type oauthAuthServerSettingsResource struct {
 type oauthAuthServerSettingsResourceModel struct {
 	Id                                          types.String `tfsdk:"id"`
 	DefaultScopeDescription                     types.String `tfsdk:"default_scope_description"`
-	Scopes                                      types.Set    `tfsdk:"scopes"`
+	Scopes                                      types.List   `tfsdk:"scopes"`
```
patrickcping, line 57: As comment above, this should probably be reverted back to `Set`

wesleymccollam: This is now reverted

PingDavidR (approved):

PR #55, opened 2023-09-25
Oauth access token manager resource

PingJamesB:


wesleymccollam (changes requested):
We should change this to be singular. For example, oauth_access_token_managers => oauth_access_token_manager

PingJamesB:
> We should change this to be singular. For example, oauth_access_token_managers => oauth_access_token_manager

Do you mean everywhere it is referenced throughout the changed files make managers => manager?

wesleymccollam:
yes

wesleymccollam (approved):

PR #56, opened 2023-09-25
remove encryptedValues

PingJamesB:


examples/resources/pingfederate_server_settings_system_keys/resource.tf
```diff
@@ -1,12 +1,28 @@
+terraform {
```
wesleymccollam, line 1: we don't need the terraform or provider blocks in the example

wesleymccollam (changes requested):

wesleymccollam (approved):

PR #60, opened 2023-09-27
Set User-Agent header in PF Terraform provider

jith-george:


internal/provider/provider.go
```diff
@@ -38,8 +39,21 @@ var (
 )
 
 // New is a helper function to simplify provider server and testing implementation.
-func New() provider.Provider {
-	return &pingfederateProvider{}
+//
```
henryrecker-pingidentity, line 42: These commented lines can be removed

henryrecker-pingidentity (approved):

henryrecker-pingidentity:
Looks good. Just be sure to include the ldflags in your goreleaser configuration when that is added.

internal/provider/provider.go
wesleymccollam: removed in latest commit

henryrecker-pingidentity (approved):

PR #67, opened 2023-10-09
validators package for schema and ValidateConfig

wesleymccollam:


internal/resource/config/oauth/oauth_auth_server_settings_resource.go
```diff
@@ -608,17 +609,12 @@ func (r *oauthAuthServerSettingsResource) ValidateConfig(ctx context.Context, re
 	if internaltypes.IsDefined(model.AllowedOrigins) {
 		aoElems := model.AllowedOrigins.Elements()
 		for _, aoElem := range aoElems {
-			aoElemString := aoElem.(basetypes.StringValue).ValueString()
-			isElemUrl := internaltypes.IsUrlFormat(aoElemString)
-			if !isElemUrl {
-				resp.Diagnostics.AddError("Invalid URL Format!", fmt.Sprintf("Please provide a valid origin. Origin \"%s\" needs to be in a valid URL-like format - \"http(s)//:<value>.<domain>\"", aoElemString))
-			}
+			validators.IsUrlFormat(aoElem, resp)
```
henryrecker-pingidentity, line 612: We could implement this as a plugin framework validator so that we could include it in the schema, since this isn't dependent on any other attributes. This way works fine as well though.

internal/resource/config/serversettings/server_settings_resource.go
```diff
@@ -860,10 +797,7 @@ func (r *serverSettingsResource) ValidateConfig(ctx context.Context, req resourc
 	// Validate contact_info email value
 	ciEmail := model.ContactInfo.Attributes()["email"]
 	if internaltypes.IsDefined(ciEmail) {
-		ciEmailStringValue := ciEmail.(types.String)
-		if internaltypes.IsNonEmptyString(ciEmailStringValue) && !internaltypes.IsEmailFormat(ciEmailStringValue.ValueString()) {
-			resp.Diagnostics.AddError("Invalid Email Format!", fmt.Sprintf("Please provide a valid email address - \"%s\" needs to be in a valid email format according to RFC 5322.  For example, \"<user>@<company>.<tld>\"", ciEmailStringValue.ValueString()))
-		}
+		validators.IsEmailFormat(ciEmail, resp)
```
henryrecker-pingidentity, line 800: Do we need these validation calls when we already have validators in the schema?

internal/resource/config/token_processor_to_token_generator_mapping_resource.go
```diff
@@ -86,6 +75,8 @@ func (r *tokenProcessorToTokenGeneratorMappingsResource) Schema(ctx context.Cont
 		},
 	}
 	AddCommonSchema(&schema)
+	AddCustomId(&schema, false, false,
+		"The id of the Token Processor to Token Generator mapping. This field is read-only and is ignored when passed in with the payload.")
```
henryrecker-pingidentity, line 79: If this field is ignored do we even need a custom_id attribute on this resource?

internal/resource/validators/validator_utils.go
```diff
@@ -0,0 +1,135 @@
...
+	//	ftp://example.com (Only http and https schemes are allowed if specified)
+	basetypesStringVal := aV.(types.String)
+	re, _ := regexp.Compile(`^(https?:\/\/)?([\da-z\.-]+)(\.[a-z]{2,6}(\.[a-z]{2,6})?)?(:[0-9]{2,5})?(\/)?(\/[^\s]*)?$`)
+	isUrl := re.MatchString(strings.ToLower(aV.(types.String).ValueString()))
+	if !internaltypes.IsNonEmptyString(basetypesStringVal) && !isUrl {
+		diag := validateConfigResp.Diagnostics
+		diag.AddError("Invalid URL Format!", fmt.Sprintf("Please provide a valid origin. Origin \"%s\" needs to be in a valid URL-like format - \"http(s)//:<value>.<domain>\"", basetypesStringVal.ValueString()))
+		return diag
+	}
+	return nil
```
henryrecker-pingidentity, line 70: Probably better to return an empty diagnostics than to return nil. Same for the following methods.

internal/resource/validators/validator_utils.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func IsEmailFormat(aV attr.Value, validateConfigResp *resource.ValidateConfigResponse) diag.Diagnostics {
```
henryrecker-pingidentity, line 73: Better to just return diagnostics here and in the next couple methods, no need to pass in the response object. The caller of this method can add to the response if necessary. Also, it's better to return an empty diagnostics objects rather than nil if there are no diagnostics to return.

henryrecker-pingidentity (changes requested):

internal/resource/configvalidators/url_validator.go
```diff
@@ -0,0 +1,43 @@
...
+
+	strVal := req.ConfigValue.ValueString()
+	isMatch, _ := regexp.MatchString(`^(https?:\/\/)`, strVal)
+	if !isMatch {
+		resp.Diagnostics.AddAttributeError(
+			req.Path,
+			"Invalid URL Format",
+			fmt.Sprintf("This %s value must start with 'http://' or 'https://'", req.ConfigValue),
+		)
+		return
```
henryrecker-pingidentity, line 37: No need for this return

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
henryrecker-pingidentity: Was it not possible to move this to the schema validator?

internal/resource/configvalidators/char_validator.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+)
+
+var _ validator.String = &charValidator{}
+
+type charValidator struct{}
+
+func (v charValidator) Description(ctx context.Context) string {
+	return "The ID must be less than 33 characters, contain no spaces, and be alphanumeric"
```
henryrecker-pingidentity, line 16: The name of this validator doesn't really align with this description. We should either make the name more specific or add some parameters to the validator to choose how many characters for example

internal/resource/configvalidators/hostname_or_ip_validator.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+)
+
+var _ validator.String = &hostnameOrIpValidator{}
+
+type hostnameOrIpValidator struct{}
+
+func (v hostnameOrIpValidator) Description(ctx context.Context) string {
+	return "Invalid hostname or IP, Please provide a valid hostname or IP address"
```
henryrecker-pingidentity, line 16: The descriptions of the validators should describe what they are validating. This sounds more like an error message.

internal/resource/configvalidators/validator_utils.go
```diff
@@ -0,0 +1,117 @@
...
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	internaltypes "github.com/pingidentity/terraform-provider-pingfederate/internal/types"
+)
+
+func IsUrlFormat(aV attr.Value, validateConfigResp *resource.ValidateConfigResponse) diag.Diagnostics {
```
henryrecker-pingidentity, line 15: Do these need to be in a utils file? They don't seem like they need to be in a shared file rather than just in the corresponding validator file

internal/resource/configvalidators/whitespace_validator.go
```diff
@@ -0,0 +1,43 @@
...
+	"regexp"
+
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+)
+
+var _ validator.String = &whitespaceValidator{}
+
+type whitespaceValidator struct{}
+
+func (v whitespaceValidator) Description(ctx context.Context) string {
```
henryrecker-pingidentity, line 15: I would rename this to imply that whitespace isn't allowed

internal/resource/config/oauth/oauth_auth_server_settings_resource.go
wesleymccollam: the recent commit has a set schema validator for this

internal/resource/configvalidators/url_validator.go
wesleymccollam: removed all non-needed returns

internal/resource/configvalidators/char_validator.go
wesleymccollam: edited this file completely with more info

internal/resource/configvalidators/hostname_or_ip_validator.go
wesleymccollam: added more validator-like descriptions rather than error-like ones

internal/resource/configvalidators/validator_utils.go
wesleymccollam: removed the utils file as it's not needed

internal/resource/configvalidators/whitespace_validator.go
wesleymccollam: renamed to noWhitespaceValidator and changed file name

internal/resource/configvalidators/url_validator.go
```diff
@@ -0,0 +1,85 @@
...
+				req.Path,
+				"Invalid URL Format Found In Set",
+				fmt.Sprintf(errorInfo, elem.(types.String).ValueString()),
+			)
+		}
+	}
+
+}
+
+func ValidateUrlsInSet() urlSetValidator {
```
henryrecker-pingidentity, line 83: nit: I think I'd use a similar name (like ValidUrlSet) here to the string one

henryrecker-pingidentity (approved):

PR #74, opened 2023-10-16
Move data source schema types

PingJamesB:
Move data source schema types to common data source package

### DIFF
--- a/internal/datasource/common/id/id_schema.go
+++ b/internal/datasource/common/id/id_schema.go
@@ -1,0 +1,18 @@
+package id
+
+import (
+	datasourceschema "github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+)
+
+func AddToDataSourceSchema(s *datasourceschema.Schema, required bool, description string) {
+	idSchemaAttr := datasourceschema.StringAttribute{}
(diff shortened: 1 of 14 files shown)


internal/datasource/common/id/id_schema.go
```diff
@@ -0,0 +1,18 @@
+package id
+
+import (
+	datasourceschema "github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+)
+
+func AddToDataSourceSchema(s *datasourceschema.Schema, required bool, description string) {
```
wesleymccollam, line 7: Could probably just be ToDataSourceSchema

internal/datasource/common/resourcelink/resource_link_schema.go
```diff
@@ -0,0 +1,19 @@
+package resourcelink
+
+import (
+	datasourceschema "github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+)
+
+func DataSourceSchema() map[string]datasourceschema.Attribute {
```
wesleymccollam, line 7: Same here, ToDataSourceSchema to match the other method naming standard

wesleymccollam (changes requested):

PR #76, opened 2023-10-18
OauthAccessTokenManager Data Source

PingJamesB:
Add OauthAccessTokenManager data source


internal/datasource/common/pluginconfiguration/plugin_configuration_schema.go
```diff
@@ -0,0 +1,143 @@
+package pluginconfiguration
+
+import (
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/listdefault"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+func ToDataSourceSchema() schema.SingleNestedAttribute {
```
henryrecker-pingidentity, line 10: I wonder if the framework has any way to convert from resource to data source schema... this is an unpleasant amount of repetition

internal/resource/config/oauth/oauth_access_token_manager_data_source.go
```diff
@@ -0,0 +1,380 @@
...
+	diags := req.Config.Get(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	apiReadOauthAccessTokenManager, httpResp, err := r.apiClient.OauthAccessTokenManagersAPI.GetTokenManager(config.ProviderBasicAuthContext(ctx, r.providerConfig), state.CustomId.ValueString()).Execute()
+
+	if err != nil {
+		if httpResp != nil && httpResp.StatusCode == 404 {
```
henryrecker-pingidentity, line 355: I don't think this special handling of 404s applies to data sources. I think a 404 in a data source should probably just return an error.

internal/datasource/common/pluginconfiguration/plugin_configuration_schema.go
PingJamesB: It doesn't seem like the framework has any solutions for this, but we could try doing through automation with the python glob package potentially. The only thing is, there are some data source schema's that do not mirror their counterpart resources very well.

henryrecker-pingidentity: Yeah that's a good point. I guess this is what we have to do for now.

internal/datasource/common/pluginconfiguration/plugin_configuration_schema.go
```diff
@@ -5,6 +5,15 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/listdefault"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+)
+
+var (
+	fieldAttrTypes = map[string]attr.Type{
+		"name":      basetypes.StringType{},
+		"value":     basetypes.StringType{},
```
henryrecker-pingidentity, line 14: Looks like this is missing encryptedValue

internal/datasource/common/pluginconfiguration/plugin_configuration_schema.go
```diff
@@ -54,6 +63,12 @@ func ToDataSourceSchema() schema.SingleNestedAttribute {
 													Optional:    false,
 													Computed:    true,
 												},
+												"encrypted_value": schema.StringAttribute{
+													Description: "For encrypted or hashed fields, this attribute contains the encrypted representation of the field's value, if a value is defined. If you do not want to update the stored value, this attribute should be passed back unchanged.",
```
henryrecker-pingidentity, line 67: You can remove the "If you do not want to update..." sentence here and below, since there's no way to update anything with a data source

internal/resource/config/oauth/oauth_access_token_manager_data_source.go
```diff
@@ -0,0 +1,380 @@
...
+	state.CustomId = types.StringValue(r.Id)
+	state.Name = types.StringValue(r.Name)
+	state.PluginDescriptorRef, respDiags = resourcelink.ToDataSourceState(ctx, &r.PluginDescriptorRef)
+	diags.Append(respDiags...)
+	state.ParentRef, respDiags = resourcelink.ToDataSourceState(ctx, r.ParentRef)
+	diags.Append(respDiags...)
+	state.Configuration, respDiags = pluginconfiguration.ToDataSourceState(configurationFromPlan, &r.Configuration)
+	diags.Append(respDiags...)
+
+	// state.AttributeContract
```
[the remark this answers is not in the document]
henryrecker-pingidentity: Was this not possible for AttributeContract?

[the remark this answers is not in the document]
PingJamesB: Oh I missed that let me see

henryrecker-pingidentity (approved):
Looks good to me 👍

PR #78, opened 2023-10-24
OauthAuthServerSettings Data Source

PingJamesB:
Add OauthAuthServerSettings data source


internal/resource/config/oauth/oauth_auth_server_settings_data_source.go
```diff
@@ -0,0 +1,620 @@
...
+
+	providerCfg := req.ProviderData.(internaltypes.ResourceConfiguration)
+	r.providerConfig = providerCfg.ProviderConfig
+	r.apiClient = providerCfg.ApiClient
+}
+
+// Read a OauthAuthServerSettingsResponse object into the model struct
+func readOauthAuthServerSettingsResponseDataSource(ctx context.Context, r *client.AuthorizationServerSettings, state *oauthAuthServerSettingsDataSourceModel, existingId *string) diag.Diagnostics {
+	var diags, respDiags diag.Diagnostics
+	state.Id = id.GenerateUUIDToState(existingId)
```
henryrecker-pingidentity, line 534: I'm wondering if there might need to be a UseStateForUnknown on the data source id attribute when we are generating a UUID. Could you check if the resource id changes on subsequent apply? We would want it to randomly generate an id on the first apply and then just maintain that value.

internal/resource/config/oauth/oauth_auth_server_settings_data_source.go
```diff
@@ -0,0 +1,620 @@
...
+	}
+
+	apiReadOauthAuthServerSettings, httpResp, err := r.apiClient.OauthAuthServerSettingsAPI.GetAuthorizationServerSettings(config.ProviderBasicAuthContext(ctx, r.providerConfig)).Execute()
+
+	if err != nil {
+		config.ReportHttpError(ctx, &resp.Diagnostics, "An error occurred while getting the OAuth Auth Server Settings", err, httpResp)
+		return
+	}
+
+	// Log response JSON
```
henryrecker-pingidentity, line 606: I believe we decided to remove these marshal json calls, because they don't really do anything at the moment

internal/resource/config/oauth/oauth_auth_server_settings_data_source.go
```diff
@@ -0,0 +1,620 @@
...
+	_, responseErr := apiReadOauthAuthServerSettings.MarshalJSON()
+	if responseErr != nil {
+		diags.AddError("There was an issue retrieving the response of OAuth Auth Server Settings: %s", responseErr.Error())
+	}
+
+	// Read the response into the state
+	id := id.GenerateUUIDStringPointer()
```
henryrecker-pingidentity, line 613: This is going to cause the id to change on every read, because we are generating a random UUID as the existing id passed in to the method below.

henryrecker-pingidentity (changes requested):

internal/resource/config/oauth/oauth_auth_server_settings_data_source.go
PingJamesB: Do you mean for just datasources or resources as well? Because they are in all of the current resources and datasources.

henryrecker-pingidentity: Ah I see we made an issue for it (1309) but haven't gone through and removed them yet. Feel free to remove this or just leave it in, since we can handle it as a part of that issue.

internal/resource/config/oauth/oauth_auth_server_settings_data_source.go
PingJamesB: /How would I do this without using a planmodifier?

henryrecker-pingidentity: Ah, yeah I forgot that the data source schema doesn't include plan modifiers.

It looks like you are already getting the state in your read method with req.Config.Get(). You can just read the value from the state struct after that call.

internal/resource/config/oauth/oauth_auth_server_settings_data_source.go
```diff
@@ -0,0 +1,620 @@
...
+	state.ClientSecretRetentionPeriod = types.Int64PointerValue(r.ClientSecretRetentionPeriod)
+	state.JwtSecuredAuthorizationResponseModeLifetime = types.Int64PointerValue(r.JwtSecuredAuthorizationResponseModeLifetime)
+	return diags
+}
+
+// Read resource information
+func (r *oauthAuthServerSettingsDataSource) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) {
+	var state oauthAuthServerSettingsDataSourceModel
+
+	diags := req.Config.Get(ctx, &state)
```
henryrecker-pingidentity, line 593: Didn't notice this before - there's no need to run this req.Config.Get because we don't need anything from the Config to retrieve this resource.

internal/resource/config/oauth/oauth_auth_server_settings_data_source.go
```diff
@@ -0,0 +1,610 @@
...
+
+	apiReadOauthAuthServerSettings, httpResp, err := r.apiClient.OauthAuthServerSettingsAPI.GetAuthorizationServerSettings(config.ProviderBasicAuthContext(ctx, r.providerConfig)).Execute()
+
+	if err != nil {
+		config.ReportHttpError(ctx, &resp.Diagnostics, "An error occurred while getting the OAuth Auth Server Settings", err, httpResp)
+		return
+	}
+
+	// Read the response into the state
+	var id = "id"
```
henryrecker-pingidentity, line 603: I'd probably use something like "oauth_auth_server_settings_id" to be specific to the data source.

henryrecker-pingidentity:
Just the one comment, looks good otherwise

henryrecker-pingidentity (approved):

PR #86, opened 2023-10-31
administrative account bug fix

wesleymccollam:


internal/acctest/administrativeaccount/administrative_account_test.go
```diff
@@ -2,111 +2,229 @@ package acctest_test
...
+			if aa.EncryptedPassword != nil {
+				encryptedPasswordVal := aa.GetEncryptedPassword()
+				passwordVal := ""
+				return encryptedPasswordVal, passwordVal
+			} else {
+				encryptedPasswordVal := ""
+				passwordVal := password
+				return encryptedPasswordVal, passwordVal
+			}
+			// return "", ""
```
henryrecker-pingidentity, line 75: nit: Can remove this line

internal/acctest/administrativeaccount/administrative_account_test.go
```diff
@@ -2,111 +2,229 @@ package acctest_test
...
+		err = acctest.TestAttributesMatchStringSlice(resourceType, &config.administrativeAccount.Username, "roles",
+			config.administrativeAccount.Roles, response.Roles)
 		if err != nil {
 			return err
 		}
+		return nil
+	}
+}
 
+func testAccCheckExpectedUpdatedAdministrativeAccountAttributes(config administrativeAccountResourceModel) resource.TestCheckFunc {
```
henryrecker-pingidentity, line 135: nit: you could combine the two check methods, using nil checks to skip the ones that aren't set in the config, to avoid some of the duplication. Up to you.

internal/resource/config/administrativeaccount/administrative_account_resource.go
```diff
@@ -75,49 +75,52 @@ func (r *administrativeAccountsResource) Schema(ctx context.Context, req resourc
...
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.UseStateForUnknown(),
-					stringplanmodifier.RequiresReplace(),
 				},
 			},
 			"password": schema.StringAttribute{
-				Description: "Password for the Account. This field is only applicable during a POST operation.",
-				Required:    true,
+				Description: "Password for the Account. This field is only applicable during account creation.",
+				Computed:    true,
```
henryrecker-pingidentity, line 96: Should this be computed?

internal/resource/config/administrativeaccount/administrative_account_resource.go
```diff
@@ -75,49 +75,52 @@ func (r *administrativeAccountsResource) Schema(ctx context.Context, req resourc
...
+				Optional:    true,
```
henryrecker-pingidentity, line 97: Can we just make this required? It is required during creation, right?

internal/resource/config/administrativeaccount/administrative_account_resource.go
```diff
@@ -75,49 +75,52 @@ func (r *administrativeAccountsResource) Schema(ctx context.Context, req resourc
...
+				Sensitive:   true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplaceIfConfigured(),
+				},
+			},
+			"encrypted_password": schema.StringAttribute{
+				Description: "Read-only attribute. This field is utilized for updating an existing Administrative Account.",
```
henryrecker-pingidentity, line 104: Might want to mention in the description that this field holds the encrypted password returned by ping fed

internal/resource/config/administrativeaccount/administrative_account_resource.go
```diff
@@ -135,26 +137,37 @@ func (r *administrativeAccountsResource) Schema(ctx context.Context, req resourc
...
 	if internaltypes.IsDefined(plan.Description) {
 		addRequest.Description = plan.Description.ValueStringPointer()
 	}
+
 	if internaltypes.IsDefined(plan.EmailAddress) {
 		addRequest.EmailAddress = plan.EmailAddress.ValueStringPointer()
 	}
-	if internaltypes.IsDefined(plan.Password) {
+
+	if internaltypes.IsDefined(plan.Password) && plan.Password.ValueString() != state.Password.ValueString() {
```
henryrecker-pingidentity, line 163: I'm not sure this check is necessary, since we already have the RequiresReplace modifier on password

internal/resource/config/administrativeaccount/administrative_account_resource.go
```diff
@@ -182,10 +196,24 @@ func (r *administrativeAccountsResource) Configure(_ context.Context, req resour
...
 }
 
-func readAdministrativeAccountResponse(ctx context.Context, r *client.AdministrativeAccount, state *administrativeAccountResourceModel, expectedValues *administrativeAccountResourceModel, passwordPlan basetypes.StringValue) {
+func readAdministrativeAccountResponse(ctx context.Context, r *client.AdministrativeAccount, state *administrativeAccountResourceModel, plan *administrativeAccountResourceModel) {
 	state.Id = types.StringValue(r.Username)
 	state.Username = types.StringValue(r.Username)
-	state.Password = types.StringValue(passwordPlan.ValueString())
+
+	// state.Password
+	if internaltypes.IsDefined(plan.Password) {
```
henryrecker-pingidentity, line 204: This if check might not be needed if we change password to required, since it will always be there in the plan.

henryrecker-pingidentity (changes requested):

internal/acctest/administrativeaccount/administrative_account_test.go
wesleymccollam: removed

internal/acctest/administrativeaccount/administrative_account_test.go
wesleymccollam: this is now combined into one method

internal/resource/config/administrativeaccount/administrative_account_resource.go
wesleymccollam: Computed removed

internal/resource/config/administrativeaccount/administrative_account_resource.go
wesleymccollam: This is now required

internal/resource/config/administrativeaccount/administrative_account_resource.go
wesleymccollam: Description has been updated

internal/resource/config/administrativeaccount/administrative_account_resource.go
wesleymccollam: This is needed for the Update since we require the password field. We can't use the UseStateForUnknown method due to the password field being added to the plan, and then request, which we don't want to happen.

internal/resource/config/administrativeaccount/administrative_account_resource.go
```diff
@@ -75,49 +75,51 @@ func (r *administrativeAccountsResource) Schema(ctx context.Context, req resourc
...
-				Description: "Password for the Account. This field is only applicable during a POST operation.",
+				Description: "Password for the Account. This field is only applicable during account creation.",
 				Required:    true,
 				Sensitive:   true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplaceIfConfigured(),
+				},
+			},
+			"encrypted_password": schema.StringAttribute{
+				Description: "Read-only attribute. This field holds the value returned from PingFederate and used for updating an existing Administrative Account.",
```
henryrecker-pingidentity, line 103: "is used"

henryrecker-pingidentity (approved):

PR #95, opened 2023-11-06
add oauth-server-settings-scopes-exclusive-scopes-datasource

PingJamesB:


internal/resource/config/oauth/oauth_auth_server_settings_scopes_exclusive_scope_data_source.go
```diff
@@ -0,0 +1,113 @@
...
+	Id          types.String `tfsdk:"id"`
+	Name        types.String `tfsdk:"name"`
+	Description types.String `tfsdk:"description"`
+	Dynamic     types.Bool   `tfsdk:"dynamic"`
+}
+
+// GetSchema defines the schema for the datasource.
+func (r *oauthAuthServerSettingsScopesExclusiveScopesDataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	schemaDef := schema.Schema{
+		Description: "Manages a OauthAuthServerSettingsScopesCommonScopes.",
```
henryrecker-pingidentity, line 42: I haven't noticed this on previous data sources, but "Manages" is probably not the right language to use for a data source, since data sources can't modify anything. I used "Describes" in the PD provider.

internal/resource/config/oauth/oauth_auth_server_settings_scopes_exclusive_scope_data_source.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+// Ensure the implementation satisfies the expected interfaces.
+var (
+	_ datasource.DataSource              = &oauthAuthServerSettingsScopesExclusiveScopesDataSource{}
+	_ datasource.DataSourceWithConfigure = &oauthAuthServerSettingsScopesExclusiveScopesDataSource{}
+)
+
+// Create a Administrative Account data source
+func NewOauthAuthServerSettingsScopesExclusiveScopesDataSource() datasource.DataSource {
```
henryrecker-pingidentity, line 22: Remove the plurals from this file, since this only gets information about a single scope - "ExclusiveScope" instead of "ExclusiveScopes"

henryrecker-pingidentity (changes requested):

henryrecker-pingidentity (approved):

PR #96, opened 2023-11-07
add Oauth Issuer Data Source BRASS-1254

PingJamesB:


internal/resource/config/oauth/issuer/oauth_issuer_data_source.go
```diff
@@ -0,0 +1,123 @@
...
+	r.apiClient = providerCfg.ApiClient
+}
+
+// Read a OauthIssuerResponse object into the model struct
+func readOauthIssuerResponseDataSource(ctx context.Context, r *client.Issuer, state *oauthIssuerDataSourceModel) {
+	//TODO why is this a pointer?
+	state.Id = types.StringValue(*r.Id)
+	state.CustomId = types.StringValue(*r.Id)
+	state.Name = types.StringValue(r.Name)
+	state.Description = types.StringValue(*r.Description)
```
henryrecker-pingidentity, line 95: Use StringPointerValue. Same for state.Path below.

internal/resource/config/oauth/issuer/oauth_issuer_data_source.go
```diff
@@ -0,0 +1,123 @@
...
+			},
+			"path": schema.StringAttribute{
+				Required: false,
+				Optional: false,
+				Computed: true,
+			},
+		},
+	}
+	id.ToDataSourceSchema(&schemaDef, false, "The persistent, unique ID for the virtual issuer. It can be any combination of [a-zA-Z0-9._-]. This property is system-assigned if not specified.")
+	id.ToDataSourceSchemaCustomId(&schemaDef, true, true, "The persistent, unique ID for the virtual issuer. It can be any combination of [a-zA-Z0-9._-]. This property is system-assigned if not specified.")
```
henryrecker-pingidentity, line 69: Could remove the last sentence here since this is required for the data source. Also, don't forget to switch from custom_id given your other review.

henryrecker-pingidentity (changes requested):

internal/resource/config/oauth/issuer/oauth_issuer_data_source.go
wesleymccollam: description updated

internal/resource/config/oauth/issuer/oauth_issuer_data_source.go
wesleymccollam: StringPointerValue's are now used

PR #98, opened 2023-11-09
custom_id to resource type id

wesleymccollam:
`custom_id` changed to resource type id. For example, the `custom_id` in the `pingfederate_authentication_policy_contract` resource is now `authentication_policy_contract_id`


henryrecker-pingidentity:
Could we consider making the id name just the last word (at least in most cases) to simplify things? For example, `adapter_id` instead of `idp_adapter_id`. And `profile_id` instead of `local_identity_identity_profile_id`?

henryrecker-pingidentity (approved):

PR #101, opened 2023-11-14
Add server-settings-general-settings-data-source

PingDavidR:
- Brass 1256


internal/resource/config/serversettings/generalsettings/server_settings_general_settings_data_source.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	providerCfg := req.ProviderData.(internaltypes.ResourceConfiguration)
+	r.providerConfig = providerCfg.ProviderConfig
+	r.apiClient = providerCfg.ApiClient
+
+}
+
+func readServerSettingsGeneralSettingsDataSource(ctx context.Context, r *client.GeneralSettings, state *serverSettingsGeneralSettingsDataSourceModel) {
+	state.Id = types.StringValue("serverSettingsGeneralSettingsId")
```
henryrecker-pingidentity, line 95: I'd use the typical terraform underscore case for this. `server_settings_general_settings_id`

internal/resource/config/serversettings/generalsettings/server_settings_general_settings_data_source.go
```diff
@@ -0,0 +1,124 @@
...
+	state.DisableAutomaticConnectionValidation = types.BoolPointerValue(r.DisableAutomaticConnectionValidation)
+	state.IdpConnectionTransactionLoggingOverride = internaltypes.StringTypeOrNil(r.IdpConnectionTransactionLoggingOverride, true)
```
henryrecker-pingidentity, line 97: We should be able to just use the `types.StringPointerValue` method for these, rather than the StringTypeOrNil method.

henryrecker-pingidentity (changes requested):

internal/resource/config/serversettings/generalsettings/server_settings_general_settings_data_source.go
PingDavidR: done

internal/resource/config/serversettings/generalsettings/server_settings_general_settings_data_source.go
PingDavidR: done

henryrecker-pingidentity (approved):

PR #107, opened 2023-11-17
CHANGELOG.md entry check for resources, dir naming cleanup

wesleymccollam:
Added logic to content check script to verify all resources have an entry in the CHANGELOG.md file.


devcheck/verifyContent.py
```diff
@@ -0,0 +1,41 @@
...
+missing_changelog_entries = []
+
+dirs = glob("./examples/resources/*/", recursive=True)
+for dir in dirs:
+  resource_name = str(dir).split("/")[-2].encode()
+  with open(r'./CHANGELOG.md', 'rb', 0) as file:
+    s = mmap.mmap(file.fileno(), 0, access=mmap.ACCESS_READ)
+    if s.find(resource_name) != -1:
+      in_changelog = True
+      continue
```
henryrecker-pingidentity, line 17: Will these continues skip the import check below?

devcheck/verifyContent.py
```diff
@@ -0,0 +1,41 @@
...
+  
+  has_import_file = os.path.isfile(f"{dir}/import.sh")
+  if not has_import_file:
+    no_import_dirs.append(dir)
+    has_no_import_files = True
+  else:
+    continue
+
+if not in_changelog:
```
henryrecker-pingidentity, line 31: Should this be checking missing_changelog_entries? You could look at the length of that instead of having a separate boolean. Same for the import checks.

docs/resources/sp_authentication_policy_contract_mapping.md
```diff
@@ -10,7 +10,57 @@ description: |-
 
 The Authentication Policy Contract (APC) to SP Adapter Mapping.
 
-
+## Example Usage
```
henryrecker-pingidentity, line 13: Good catch on renaming the example to fix this

devcheck/verifyContent.py
```diff
@@ -0,0 +1,41 @@
...
+for dir in dirs:
+  resource_name = str(dir).split("/")[-2].encode()
+  with open(r'./CHANGELOG.md', 'rb', 0) as file:
+    s = mmap.mmap(file.fileno(), 0, access=mmap.ACCESS_READ)
+    if s.find(resource_name) != -1:
+      in_changelog = True
+      continue
+    else:
+      in_changelog = False
+      if resource_name not in missing_changelog_entries:
```
henryrecker-pingidentity, line 20: You could make missing_changelog_entries a set so you don't have to check before appending

henryrecker-pingidentity (approved):

PR #111, opened 2023-11-20
oauth client resource

wesleymccollam:


examples/resources/pingfederate_oauth_client/resource.tf
```diff
@@ -0,0 +1,80 @@
+resource "pingfederate_oauth_client" "simpleOauthClient" {
```
henryrecker-pingidentity, line 1: Do we need this one in the example?

internal/acctest/oauthclient/oauth_client_test.go
```diff
@@ -0,0 +1,460 @@
...
+		oauthClientId,
+		acctest.StringSliceToTerraformString(resourceModel.grantTypes),
+		resourceModel.name,
+		optionalRedirectUris,
+		optionalHcl,
+	)
+}
+
+// Test that the expected attributes are set on the PingFederate server
+func testAccCheckExpectedOauthClientAttributes(config oauthClientResourceModel, testOptionalAttributes bool) resource.TestCheckFunc {
```
henryrecker-pingidentity, line 277: You could read the "includeOptionalAttributes" property from the model rather than using a separate boolean parameter in this function.

internal/resource/config/administrativeaccount/administrative_account_data_source.go
```diff
@@ -162,14 +161,6 @@ func (r *administrativeAccountDataSource) Read(ctx context.Context, req datasour
 		return
 	}
 
-	// Log response JSON
```
henryrecker-pingidentity, line 165: Good catch

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -0,0 +1,1579 @@
...
+				Description: "Allows an administrator to override the Persistent Grant Lifetime set globally for the OAuth AS. Defaults to SERVER_DEFAULT.",
+				Computed:    true,
+				Optional:    true,
+				Default:     stringdefault.StaticString("SERVER_DEFAULT"),
+			},
+			"persistent_grant_expiration_time": schema.Int64Attribute{
+				Description: "The persistent grant expiration time. -1 indicates an indefinite amount of time.",
+				Computed:    true,
+				Optional:    true,
+				Default:     int64default.StaticInt64(0),
```
henryrecker-pingidentity, line 281: Should this default be -1? I assume 0 would make things immediately expire, which seems a little weird.

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -0,0 +1,1579 @@
...
+			"persistent_grant_reuse_type": schema.StringAttribute{
+				Description: "Allows and administrator to override the Reuse Existing Persistent Access Grants for Grant Types set globally for OAuth AS. Defaults to SERVER_DEFAULT",
+				Computed:    true,
+				Optional:    true,
+				Default:     stringdefault.StaticString("SERVER_DEFAULT"),
+				Validators: []validator.String{
+					stringvalidator.OneOf("SERVER_DEFAULT", "OVERRIDE_SERVER_DEFAULT"),
+				},
+			},
+			"persistent_grant_reuse_grant_types": schema.SetAttribute{
```
henryrecker-pingidentity, line 322: Is there a default for this?

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -0,0 +1,1579 @@
...
+						Validators: []validator.String{
+							configvalidators.ValidUrl(),
+						},
+					},
+				},
+				PlanModifiers: []planmodifier.Object{
+					objectplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"client_auth": schema.SingleNestedAttribute{
```
henryrecker-pingidentity, line 545: Is there a default for this attribute?

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -0,0 +1,1579 @@
...
+	var model oauthClientResourceModel
+	resp.Diagnostics.Append(req.Config.Get(ctx, &model)...)
+
+	// Persistent Grant Expiration Validation
+	if internaltypes.IsDefined(model.PersistentGrantExpirationType) && model.PersistentGrantExpirationType.ValueString() != "OVERRIDE_SERVER_DEFAULT" {
+		if !internaltypes.IsDefined(model.PersistentGrantExpirationTime) || !internaltypes.IsDefined(model.PersistentGrantExpirationTimeUnit) {
+			resp.Diagnostics.AddError("persistent_grant_expiration_type must be configured to \"OVERRIDE_SERVER_DEFAULT\" to modify the other persistent_grant_expiration values.", "")
+		}
+	}
+	if (internaltypes.IsDefined(model.PersistentGrantExpirationTime) || internaltypes.IsDefined(model.PersistentGrantExpirationTimeUnit)) && model.PersistentGrantExpirationType.ValueString() != "OVERRIDE_SERVER_DEFAULT" {
```
henryrecker-pingidentity, line 916: What is the difference between this check and the one directly above?

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -0,0 +1,1579 @@
...
+		resp.Diagnostics.AddError("persistent_grant_expiration_type must be configured to \"OVERRIDE_SERVER_DEFAULT\" to modify the other persistent_grant_expiration values.", "")
+	}
+
+	// Refresh Token Rolling Validation
+	if model.RefreshTokenRollingIntervalType.ValueString() == "OVERRIDE_SERVER_DEFAULT" && !internaltypes.IsDefined(model.RefreshTokenRollingInterval) {
+		resp.Diagnostics.AddError("refresh_token_rolling_interval must be configured when refresh_token_rolling_interval_type is \"OVERRIDE_SERVER_DEFAULT\".", "")
+	}
+	if internaltypes.IsDefined(model.RefreshTokenRollingInterval) && model.RefreshTokenRollingIntervalType.ValueString() != "OVERRIDE_SERVER_DEFAULT" {
```
henryrecker-pingidentity, line 924: These also look the same. You could do a boolean comparison here potentially to avoid repeating it. Something like `if (model.RefreshTokenRollingIntervalType.ValueString() == "OVERRIDE_SERVER_DEFAULT") != internaltypes.IsDefined(model.RefreshTokenRollingInterval)`

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -0,0 +1,1579 @@
...
+					resp.Diagnostics.AddError("jwks_settings must be defined when client_auth is configured to \"PRIVATE_KEY_JWT\".", "")
+				}
+			case "CERTIFICATE":
+				if !internaltypes.IsDefined(clientAuthAttributes["client_cert_subject_dn"]) || !internaltypes.IsDefined(clientAuthAttributes["client_cert_issuer_dn"]) {
+					resp.Diagnostics.AddError("client_cert_subject_dn and client_cert_issuer_dn must be defined when client_auth is configured to \"CERTIFICATE\".", "")
+				}
+			}
+		}
+	} else {
+		clientAuthDefined = false
```
henryrecker-pingidentity, line 948: You can set `clientAuthDefined = internaltypes.IsDefined(model.ClientAuth)` rather than using an if-else

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -0,0 +1,1579 @@
...
+	// grant_types is required, don't need nil check here
+	var hasCibaGrantType bool
+	for _, grantType := range model.GrantTypes.Elements() {
+		grantTypeVal := grantType.(types.String).ValueString()
+		if grantTypeVal == "CLIENT_CREDENTIALS" {
+			if clientAuthDefined {
+				clientAuthType := clientAuthAttributes["type"].(types.String).ValueString()
+				clientAuthSecret := clientAuthAttributes["secret"].(types.String).ValueString()
+				if clientAuthType != "NONE" && clientAuthSecret == "" {
+					resp.Diagnostics.AddError("client_auth must be defined when \"CLIENT_CREDENTIALS\" is included in grant_types.", "")
```
henryrecker-pingidentity, line 961: This error message doesn't seem right

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -0,0 +1,1579 @@
...
+	if !hasCibaGrantType && (internaltypes.IsDefined(model.CibaDeliveryMode) ||
+		internaltypes.IsDefined(model.CibaNotificationEndpoint) ||
+		internaltypes.IsDefined(model.CibaPollingInterval) ||
+		internaltypes.IsDefined(model.CibaRequireSignedRequests) ||
+		internaltypes.IsDefined(model.CibaRequestObjectSigningAlgorithm) ||
+		internaltypes.IsDefined(model.CibaUserCodeSupported)) {
+		resp.Diagnostics.AddError("ciba attributes can only be configured when \"CIBA\" is included in grant_types.", "")
+	}
+	if hasCibaGrantType &&
+		(internaltypes.IsDefined(model.CibaDeliveryMode) &&
```
henryrecker-pingidentity, line 982: You shouldn't need this IsDefined check, since ValueString will just return "" if something isn't defined.

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -0,0 +1,1579 @@
...
+		oidcPolicySectorIdentifierUri := oidcPolicy["sector_identifier_uri"]
+		if (pairwiseIdentifierUserType != nil && !pairwiseIdentifierUserType.(types.Bool).ValueBool()) && internaltypes.IsDefined(oidcPolicySectorIdentifierUri) {
+			resp.Diagnostics.AddError("sector_identifier_uri can only be configured when pairwise_identifier_user_type is set to true.", "")
+		}
+	}
+
+	// JWKS Settings Validation
+	if !internaltypes.IsDefined(model.JwksSettings) {
+		if internaltypes.IsDefined(model.TokenIntrospectionEncryptionAlgorithm) {
+			resp.Diagnostics.AddError("token_introspection_encryption_algorithm must not be configured jwks_settings is configured.", "")
```
henryrecker-pingidentity, line 1033: This message seems wrong - if we get to here, then we've confirmed that jwks_settings is *not* configured

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -0,0 +1,1579 @@
...
+	state.DefaultAccessTokenManagerRef, respDiags = resourcelink.ToState(ctx, r.DefaultAccessTokenManagerRef)
+	diags.Append(respDiags...)
+	state.RestrictToDefaultAccessTokenManager = types.BoolPointerValue(r.RestrictToDefaultAccessTokenManager)
+	state.ValidateUsingAllEligibleAtms = types.BoolPointerValue(r.ValidateUsingAllEligibleAtms)
+	state.RefreshRolling = types.StringPointerValue(r.RefreshRolling)
+	state.RefreshTokenRollingIntervalType = types.StringPointerValue(r.RefreshTokenRollingIntervalType)
+	state.RefreshTokenRollingInterval = types.Int64PointerValue(r.RefreshTokenRollingInterval)
+	state.PersistentGrantExpirationType = types.StringPointerValue(r.PersistentGrantExpirationType)
+	state.PersistentGrantExpirationTime = types.Int64PointerValue(r.PersistentGrantExpirationTime)
+	if types.StringPointerValue(r.PersistentGrantExpirationTimeUnit).ValueString() == "" {
```
henryrecker-pingidentity, line 1062: Could this instead be a nil check? `if r.PersistentGrantExpirationTimeUnit == nil`?

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -0,0 +1,1579 @@
...
+		if secondarySecretsFromPlan != nil && len(secondarySecretsFromPlan.(types.Set).Elements()) > 0 {
+			for _, secondarySecretsFromPlan := range clientAuthFromPlan["secondary_secrets"].(types.Set).Elements() {
+				secondarySecretsAttrVal, respDiags := types.ObjectValueFrom(ctx, secondarySecretsAttrType, secondarySecretsFromPlan)
+				diags.Append(respDiags...)
+				secondarySecretsSetSlice = append(secondarySecretsSetSlice, secondarySecretsAttrVal)
+			}
+			secondarySecretsObjToState, respDiags = types.SetValue(types.ObjectType{AttrTypes: secondarySecretsAttrType}, secondarySecretsSetSlice)
+			diags.Append(respDiags...)
+		} else {
+			secondarySecretsObjToState = types.SetNull(types.ObjectType{AttrTypes: secondarySecretsAttrType})
```
henryrecker-pingidentity, line 1115: Should this be an empty set rather than a null set?

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -0,0 +1,1579 @@
...
+
+	if internaltypes.IsDefined(plan.DefaultAccessTokenManagerRef) {
+		addRequest.DefaultAccessTokenManagerRef = client.NewResourceLinkWithDefaults()
+		err := json.Unmarshal([]byte(internaljson.FromValue(plan.DefaultAccessTokenManagerRef, false)), addRequest.DefaultAccessTokenManagerRef)
+		if err != nil {
+			return err
+		}
+	}
+
+	if internaltypes.IsDefined(plan.RestrictToDefaultAccessTokenManager) {
```
henryrecker-pingidentity, line 1226: Not 100% sure but I don't think the IsDefined checks are necessary when we aren't using the Json Unmarshal method. If it's not defined, the ValueXPointer method should return nil, and that's what we'd want.

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -0,0 +1,1579 @@
...
+		if httpResp != nil && httpResp.StatusCode == 404 {
+			config.ReportHttpErrorAsWarning(ctx, &resp.Diagnostics, "An error occurred while getting the OauthClient", err, httpResp)
+			resp.State.RemoveResource(ctx)
+		} else {
+			config.ReportHttpError(ctx, &resp.Diagnostics, "An error occurred while getting the  OauthClient", err, httpResp)
+		}
+	}
+
+	// Read the response into the state
+	readOauthClientResponse(ctx, apiReadOauthClient, &state, &state)
```
henryrecker-pingidentity, line 1518: Need to handle diags here

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -0,0 +1,1579 @@
...
+	var state oauthClientResourceModel
+	diags = readOauthClientResponse(ctx, updateOauthClientResponse, &plan, &state)
+	resp.Diagnostics.Append(diags...)
+
+	// Update computed values
+	diags = resp.State.Set(ctx, plan)
+	resp.Diagnostics.Append(diags...)
+}
+
+// This config object is edit-only, so Terraform can't delete it.
```
henryrecker-pingidentity, line 1560: Remove this comment

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -0,0 +1,1579 @@
...
+	diags := req.Plan.Get(ctx, &plan)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	createOauthClient := client.NewClient(plan.ClientId.ValueString(), grantTypes(plan.GrantTypes), plan.Name.ValueString())
+	err := addOptionalOauthClientFields(ctx, createOauthClient, plan)
+	if err != nil {
+		resp.Diagnostics.AddError("Failed to add optional properties to add request for OauthClient", err.Error())
```
henryrecker-pingidentity, line 1477: You could change these error messages from "OauthClient" to "OAuth Client"

henryrecker-pingidentity (changes requested):

examples/resources/pingfederate_oauth_client/resource.tf
wesleymccollam: I thought it would be nice if a user just wanted all of the default values and a simple example to have.

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: If you don't send a value, this value is 0, by default

examples/resources/pingfederate_oauth_client/resource.tf
henryrecker-pingidentity: I think most of our examples have pretty much fully fleshed-out realistic versions of these resources, so imo we don't really need the simple one.

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: The default value is for the type equal to NONE, and the secret is empty. So if it's not different, this would mean it's not set properly.

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: I got the message based on copilot, I will reword it

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: This value is default and isn't returned by PingFed if unchanged. I'll add a better check though.

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: I'll change this to an empty set.

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: Found this is a few other resources as well. This will be included in the next commit

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: There's not. It's not always included in the request or returned from the client.

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: re-worded

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: updated check to `if r.GetPersistentGrantExpirationTimeUnit() == ""`

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: done

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: these are all updated besides sets and objects

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: done

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: updated

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: descriptions updated

internal/resource/config/oauth/client/oauth_client_resource.go
henryrecker-pingidentity: I was confused on this one because in this if statement `client_auth` is defined, and the error message is implying that it isn't

internal/resource/config/oauth/client/oauth_client_resource.go
henryrecker-pingidentity: Looks like this is still using SetNull

wesleymccollam: whoops! will get this added.

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: ahh, I see. I'll update this description.

internal/resource/config/oauth/client/oauth_client_data_source.go
```diff
@@ -0,0 +1,764 @@
...
+	return &oauthClientDataSource{}
+}
+
+// oauthClientDataSource is the datasource implementation.
+type oauthClientDataSource struct {
+	providerConfig internaltypes.ProviderConfiguration
+	apiClient      *client.APIClient
+}
+
+type oauthClientDataSourceModel struct {
```
henryrecker-pingidentity, line 56: If this is the same as the resource model, we could just use that instead of redeclaring it here

docs/data-sources/oauth_client.md
```diff
@@ -0,0 +1,330 @@
...
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `client_id` (String) A unique identifier the client provides to the Resource Server to identify itself. This identifier is included with every request the client makes. For PUT requests, this field is optional and it will be overridden by the 'id' parameter of the PUT request.
+
+### Optional
```
henryrecker-pingidentity, line 28: These should presumably all be read-only

wesleymccollam: oops.. nice catch

wesleymccollam: fixed

internal/resource/config/oauth/client/oauth_client_data_source.go
wesleymccollam: changed name to `oauthClientModel` and referenced in both resource and data source.

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: default added

henryrecker-pingidentity (approved):

PR #113, opened 2023-11-20
server settings captcha settings acc test coverage

wesleymccollam:


internal/acctest/serversettings/server_settings_test.go
```diff
@@ -252,6 +269,14 @@ func testAccCheckExpectedServerSettingsAttributes(config serverSettingsResourceM
 			}
 		}
 
+		if config.captchaSettings != nil {
+			err = acctest.TestAttributesMatchString(resourceType, nil, "site_key",
```
henryrecker-pingidentity, line 273: Does PF even return this? I assumed it would be empty since it's a "key" of some sort

henryrecker-pingidentity (approved):
Thanks for updating this.

internal/acctest/serversettings/server_settings_test.go
wesleymccollam: It returns siteKey and encryptedSecretKey, just not secretKey - like the password behavior.

PR #114, opened 2023-11-21
Refactor server log settings resource to match data source

PingDavidR:
- closes BRASS-1340


CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 # v0.3.0 (unreleased)
+### ENHANCEMENTS
```
henryrecker-pingidentity, line 2: We don't need this in the changelog, since there's no behavior change

henryrecker-pingidentity (changes requested):

henryrecker-pingidentity (approved):

PR #116, opened 2023-11-21
redirect validation data source

wesleymccollam:
- redirect validation data source
- state clean up in resource file


henryrecker-pingidentity (approved):
Looks good, thanks for cleaning up the resource too

PR #120, opened 2023-11-27
token processor to token generator mapping data source

wesleymccollam:


internal/datasource/common/attributesources/attribute_sources_state.go
```diff
@@ -0,0 +1,167 @@
+package attributesources
```
henryrecker-pingidentity, line 1: Is this necessary? It doesn't seem to be using any datasource-specific logic, so can we just use the existing common logic for attribute_sources?

Generally the schema types for data sources seem to be different, but outside of schema I'm not sure this is necessary.

internal/resource/config/tokenprocessortotokengeneratormapping/token_processor_to_token_generator_mapping_data_source.go
```diff
@@ -0,0 +1,110 @@
...
+
+	diags := req.Config.Get(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	apiReadTokenProcessorToTokenGeneratorMapping, httpResp, err := r.apiClient.TokenProcessorToTokenGeneratorMappingsAPI.GetTokenToTokenMappingById(config.ProviderBasicAuthContext(ctx, r.providerConfig), state.Id.ValueString()).Execute()
+
+	if err != nil {
+		if httpResp != nil && httpResp.StatusCode == 404 {
```
henryrecker-pingidentity, line 95: Don't need 404 logic in data sources

henryrecker-pingidentity (changes requested):

henryrecker-pingidentity (approved):
Looks good. Thanks for making fixes in the other data sources for this.

PR #121, opened 2023-11-28
openidconnect policies data source

PingDavidR:
- closes Brass-1297


examples/data-sources/pingfederate_oauth_open_id_connect_policy/data-source.tf
```diff
@@ -0,0 +1,2 @@
+data "pingfederate_oauth_open_id_connect_policy" "myOauthOIDCPolicyExample" {
```
henryrecker-pingidentity, line 1: Missing policy_id from the example

henryrecker-pingidentity (changes requested):

examples/data-sources/pingfederate_oauth_open_id_connect_policy/data-source.tf
```diff
@@ -0,0 +1,3 @@
...
+  policy_id = pingfederate_oauth_open_id_connect_policy.myOauthOIDCPolicyExample.id
```
henryrecker-pingidentity, line 2: I'd just put some string here, since there's no other resource in this example

henryrecker-pingidentity (approved):

PR #123, opened 2023-11-29
data store data source

wesleymccollam:


examples/data-sources/pingfederate_data_store/data-source.tf
```diff
@@ -0,0 +1,3 @@
+data "pingfederate_data_store" "myDataStore" {
+  data_store_id = pingfederate_data_store.myDataStore.data_store_id
```
henryrecker-pingidentity, line 2: Can just use a string here

internal/resource/config/datastore/data_store_data_source.go
```diff
@@ -0,0 +1,105 @@
...
+
+type dataStoreDataSource struct {
+	providerConfig internaltypes.ProviderConfiguration
+	apiClient      *client.APIClient
+}
+
+// GetSchema defines the schema for the datasource.
+func (r *dataStoreDataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	schema := schema.Schema{
+		Description: "Describes a data store datasource",
```
henryrecker-pingidentity, line 31: Should be "Describes a data store"

internal/resource/config/datastore/jdbc_data_store.go
```diff
@@ -229,19 +342,56 @@ func toStateJdbcDataStore(con context.Context, jdbcDataStore *client.JdbcDataSto
...
+
+	if jdbcDataStore == nil {
+		diags.AddError("Failed to read JDBC data store from PingFederate.", "The response from PingFederate was nil.")
+		return types.ObjectNull(jdbcDataStoreAttrType), diags
+	}
+
+	connectionUrlSetVal, diags := types.SetValueFrom(con, jdbcTagConfigAttrType, jdbcDataStore.ConnectionUrlTags)
+	allDiags = append(allDiags, diags...)
+
+	jdbcAttrValue := map[string]attr.Value{
```
henryrecker-pingidentity, line 356: If you want to slightly reduce repetition, you could put most of this into a common method. Since from what I can tell it's just the password/encrypted_password that is different

henryrecker-pingidentity (changes requested):
Only a few small comments

examples/data-sources/pingfederate_data_store/data-source.tf
wesleymccollam: updated

internal/resource/config/datastore/data_store_data_source.go
wesleymccollam: updated

internal/resource/config/datastore/jdbc_data_store.go
wesleymccollam: added methods to internaltypes package for this

internal/types/utils.go
```diff
@@ -325,3 +325,22 @@ func CheckListKeyMatch(k string, list []string) bool {
 	}
 	return false
 }
+
+// Add the password and encrypted_password to existing map[string]attr.Type
+func AddPasswordToMapStringAttrType(mapStringAttrType map[string]attr.Type) map[string]attr.Type {
```
henryrecker-pingidentity, line 330: You could combine these into one generic method. Just make the name of the attribute a parameter

PR #126, opened 2023-12-01
password credential validator validate config

wesleymccollam:


internal/resource/config/passwordcredentialvalidator/password_credential_validator_resource.go
```diff
@@ -122,6 +122,119 @@ func (r *passwordCredentialValidatorResource) Schema(ctx context.Context, req re
...
+
+func (r *passwordCredentialValidatorResource) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
+	var model passwordCredentialValidatorModel
+	resp.Diagnostics.Append(req.Config.Get(ctx, &model)...)
+
+	tableNames := []string{}
+	nestedTableFieldNameValues := []string{}
+	var rowCount int
+	configuration := model.Configuration.Attributes()
+	tables := configuration["tables"].(types.List).Elements()
```
henryrecker-pingidentity, line 149: Are configuration and configuration.tables required in the schema? Are we sure they will be set here?

internal/types/utils.go
```diff
@@ -67,6 +67,17 @@ func StringSliceContains(slice []string, value string) bool {
 	return false
 }
 
+// Checks value occurs the expected number of times in the slice
+func ValidateStringCountInSlice(slice []string, value string, expectedCount int) (bool, int) {
```
henryrecker-pingidentity, line 71: I think this function would make a little more sense as just a counting function, that just returns the number of times that string was found. It's easier to understand reading the code if this just returns a number and wherever it is called does the check for if that equals zero or whatever.

If you switch to using a map then you may not need this method at all.

internal/resource/config/passwordcredentialvalidator/password_credential_validator_resource.go
```diff
@@ -122,6 +122,119 @@ func (r *passwordCredentialValidatorResource) Schema(ctx context.Context, req re
...
+		fieldName := field["name"].(types.String).ValueString()
+		fieldNameValues = append(fieldNameValues, fieldName)
+	}
+
+	pluginDescriptorRefId := model.PluginDescriptorRef.Attributes()["id"].(types.String).ValueString()
+	switch pluginDescriptorRefId {
+	case "com.pingconnect.alexandria.pingfed.pcv.PingOnePasswordValidator":
+		hasClientId, clientIdCount := internaltypes.ValidateStringCountInSlice(fieldNameValues, "Client Id", 1)
+		if !hasClientId || clientIdCount == 0 {
+			resp.Diagnostics.AddError("The \"Client Id\" field is required for the PingOne for Enterprise Directory Password Credential Validator", "")
```
henryrecker-pingidentity, line 182: It seems like these errors should indicate exactly one value is expected. Since you would get this same error message if you had two Client Id values, if I'm understanding it right.

internal/resource/config/passwordcredentialvalidator/password_credential_validator_resource.go
```diff
@@ -122,6 +122,119 @@ func (r *passwordCredentialValidatorResource) Schema(ctx context.Context, req re
...
+				for _, field := range fields {
+					fieldRow := field.(types.Object).Attributes()
+					nestedTableFieldName := fieldRow["name"].(types.String).ValueString()
+					nestedTableFieldNameValues = append(nestedTableFieldNameValues, nestedTableFieldName)
+				}
+			}
+		}
+	}
+
+	fieldNameValues := []string{}
```
henryrecker-pingidentity, line 169: If all you ever do with these slices is check for existence of an element, I'd use a map[string]bool instead, so you don't have to iterate over the whole slice every time.

internal/resource/config/passwordcredentialvalidator/password_credential_validator_resource.go
```diff
@@ -122,6 +122,119 @@ func (r *passwordCredentialValidatorResource) Schema(ctx context.Context, req re
...
+	case "org.sourceid.saml20.domain.RadiusUsernamePasswordCredentialValidator":
+		hasRadiusDs, radiusDsCount := internaltypes.ValidateStringCountInSlice(tableNames, "RADIUS Servers", 1)
+		if !hasRadiusDs || radiusDsCount == 0 {
+			resp.Diagnostics.AddError("At least one \"RADIUS Servers\" table is required for the RADIUS Username Password Credential Validator", "")
+		}
+
+	case "org.sourceid.saml20.domain.SimpleUsernamePasswordCredentialValidator":
+		hasCorrectUsernameCount, usernameCount := internaltypes.ValidateStringCountInSlice(nestedTableFieldNameValues, "Username", rowCount)
+		if !hasCorrectUsernameCount || usernameCount == 0 {
+			resp.Diagnostics.AddError("The \"Username\" field is required for the Simple Username Password Credential Validator", "")
```
henryrecker-pingidentity, line 224: These messages that are specifically looking for these fields within tables should probably be clear about that. Right now they use the same language as the checks for fields outside of tables.

internal/resource/config/passwordcredentialvalidator/password_credential_validator_resource.go
```diff
@@ -122,6 +122,119 @@ func (r *passwordCredentialValidatorResource) Schema(ctx context.Context, req re
...
+		}
+
+	case "org.sourceid.saml20.domain.RadiusUsernamePasswordCredentialValidator":
+		hasRadiusDs, radiusDsCount := internaltypes.ValidateStringCountInSlice(tableNames, "RADIUS Servers", 1)
+		if !hasRadiusDs || radiusDsCount == 0 {
+			resp.Diagnostics.AddError("At least one \"RADIUS Servers\" table is required for the RADIUS Username Password Credential Validator", "")
+		}
+
+	case "org.sourceid.saml20.domain.SimpleUsernamePasswordCredentialValidator":
+		hasCorrectUsernameCount, usernameCount := internaltypes.ValidateStringCountInSlice(nestedTableFieldNameValues, "Username", rowCount)
```
henryrecker-pingidentity, line 222: For these ones that are expected certain fields in every single row, I might just move that check up to the part where you are looping through them.

henryrecker-pingidentity (changes requested):

internal/resource/config/passwordcredentialvalidator/password_credential_validator_resource.go
wesleymccollam: you can only have one value of each in this one

henryrecker-pingidentity (approved):

PR #131, opened 2023-12-06
multiple terraform version acceptance test coverage

wesleymccollam:


henryrecker-pingidentity (approved):
If the tests eventually take a long time, we ought to consider moving the multi-version tests to a cron system so we don't have to wait for three runs every pull request.

PR #139, opened 2023-12-11
add missing common packages

wesleymccollam:


internal/resource/config/localidentity/identityprofile/local_identity_identity_profile_common.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	client "github.com/pingidentity/pingfederate-go-client/v1125/configurationapi"
+	"github.com/pingidentity/terraform-provider-pingfederate/internal/resource/common/resourcelink"
+)
+
+var (
+	authSourcesAttrTypes = map[string]attr.Type{
+		"id":     basetypes.StringType{},
```
henryrecker-pingidentity, line 16: This may be better for another ticket, but I think we ought to be using `types.StringType` (and similar for other types) rather than the `basetypes` version. By doing it this way with `basetypes`, we are instantiating a new struct for every one of these attributes. `types.StringType` avoids that and uses the same existing struct for each occurrence.

internal/resource/config/localidentity/identityprofile/local_identity_identity_profile_common.go
```diff
@@ -0,0 +1,176 @@
...
+	RegistrationConfig      types.Object `tfsdk:"registration_config"`
+	ProfileConfig           types.Object `tfsdk:"profile_config"`
+	FieldConfig             types.Object `tfsdk:"field_config"`
+	EmailVerificationConfig types.Object `tfsdk:"email_verification_config"`
+	DataStoreConfig         types.Object `tfsdk:"data_store_config"`
+	ProfileEnabled          types.Bool   `tfsdk:"profile_enabled"`
+}
+
+// Read a DseeCompatAdministrativeAccountResponse object into the model struct
+func readLocalIdentityIdentityProfileResponseDataSource(ctx context.Context, r *client.LocalIdentityProfile, state *localIdentityIdentityProfileModel) diag.Diagnostics {
```
henryrecker-pingidentity, line 111: Should this be in common? I see the method still has the "DataSource" suffix, and I'm not sure it's used in the resource.

internal/resource/config/oauth/authserversettings/oauth_auth_server_settings_resource.go
```diff
@@ -40,24 +38,19 @@ var (
 	_ resource.Resource                = &oauthAuthServerSettingsResource{}
 	_ resource.ResourceWithConfigure   = &oauthAuthServerSettingsResource{}
 	_ resource.ResourceWithImportState = &oauthAuthServerSettingsResource{}
+)
 
+var (
```
henryrecker-pingidentity, line 43: This doesn't need to be a separate block

internal/resource/config/localidentity/identityprofile/local_identity_identity_profile_common.go
wesleymccollam: I have added another case to handle this.

henryrecker-pingidentity (approved):

henryrecker-pingidentity:
Looks good 👍

PR #144, opened 2023-12-12
hardcoded filedata to env vars

wesleymccollam:


henryrecker-pingidentity:
I'd add a check in the test to fail early if those variables aren't set at all, informing that you need to set them

henryrecker-pingidentity (approved):

PR #150, opened 2023-12-20
Authentication Policies Settings Resource and Data Source

PingDavidR:
- closes PDI-1377


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+# v0.5.0 (unreleased)
+### FEATURES
+#### Resources
+* **New Resource:** `pingfederate_authentication_policies_settings` ([#150](https://github.com/pingidentity/terraform-provider-pingfederate/pull/150))
+
+#### Data Sources
+* **New Data Source:** `pingfederate_authentication_policies_settings` ([#T150](https://github.com/pingidentity/terraform-provider-pingfederate/pull/150))
```
henryrecker-pingidentity, line 7: Misplaced T here

examples/resources/pingfederate_authentication_policies_settings/import.sh
```diff
@@ -0,0 +1,2 @@
+# "authenticationPoliciesSettingsId" should be "id"
```
henryrecker-pingidentity, line 1: This is probably best for another issue, and I know this is copied from our other examples, but this comment doesn't seem accurate. In fact with resources like this that authenticationPoliciesSettingsId value shouldn't even matter, no matter what the user supplies.

internal/acctest/authenticationpoliciessettings/authentication_policies_settings_test.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() { acctest.ConfigurationPreCheck(t) },
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
+			"pingfederate": providerserver.NewProtocol6WithError(provider.NewTestProvider()),
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAuthenticationPoliciesSettings(resourceName, initialResourceModel),
```
henryrecker-pingidentity, line 39: To align with our other tests, the first step should set only the required properties of the resource. In this case there are no required properties, so the first step can just apply some HCL with no attributes in the resource itself. 
There should also be a 4th test step that returns to the initial empty model.

internal/resource/config/authenticationpolicies/settings/authentication_policies_settings_common.go
```diff
@@ -0,0 +1,25 @@
...
+	EnableSpAuthnSelection  types.Bool   `tfsdk:"enable_sp_authn_selection"`
+}
+
+func readAuthenticationPoliciesSettingsResponse(ctx context.Context, r *client.AuthenticationPoliciesSettings, state *authenticationPoliciesSettingsModel, existingId *string) {
+	if existingId != nil {
+		state.Id = types.StringValue(*existingId)
+	} else {
+		state.Id = id.GenerateUUIDToState(existingId)
+	}
+	state.EnableIdpAuthnSelection = types.BoolValue(*r.EnableIdpAuthnSelection)
```
henryrecker-pingidentity, line 23: Use types.BoolPointerValue here and on the next line, rather than dereferencing the pointer

internal/resource/config/authenticationpolicies/settings/authentication_policies_settings_data_source.go
```diff
@@ -0,0 +1,95 @@
...
+	providerConfig internaltypes.ProviderConfiguration
+	apiClient      *client.APIClient
+}
+
+// GetSchema defines the schema for the datasource.
+func (r *authenticationPoliciesSettingsDataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	schema := schema.Schema{
+		Description: "Describes the authentication policies settings",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
```
henryrecker-pingidentity, line 37: The id.ToDataSourceSchema call below adds `id` to the schema, so we don't need to duplicate it here

internal/resource/config/authenticationpolicies/settings/authentication_policies_settings_common.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/pingidentity/terraform-provider-pingfederate/internal/resource/common/id"
+)
+
+type authenticationPoliciesSettingsModel struct {
+	Id                      types.String `tfsdk:"id"`
+	EnableIdpAuthnSelection types.Bool   `tfsdk:"enable_idp_authn_selection"`
+	EnableSpAuthnSelection  types.Bool   `tfsdk:"enable_sp_authn_selection"`
+}
+
+func readAuthenticationPoliciesSettingsResponse(ctx context.Context, r *client.AuthenticationPoliciesSettings, state *authenticationPoliciesSettingsModel, existingId *string) {
```
henryrecker-pingidentity, line 17: Looks like `ctx` isn't used, so you can remove it from the parameters

internal/resource/config/authenticationpolicies/settings/authentication_policies_settings_resource.go
```diff
@@ -0,0 +1,227 @@
...
+	providerConfig internaltypes.ProviderConfiguration
+	apiClient      *client.APIClient
+}
+
+// GetSchema defines the schema for the resource.
+func (r *authenticationPoliciesSettingsResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	schema := schema.Schema{
+		Description: "Manages Authentication Policies Settings",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
```
henryrecker-pingidentity, line 40: Same as the data source - no need to redefine id here

internal/resource/config/authenticationpolicies/settings/authentication_policies_settings_resource.go
```diff
@@ -0,0 +1,227 @@
...
+		},
+	}
+
+	id.ToSchema(&schema)
+	resp.Schema = schema
+}
+
+func addOptionalAuthenticationPoliciesSettingsFields(ctx context.Context, addRequest *client.AuthenticationPoliciesSettings, plan authenticationPoliciesSettingsModel) error {
+
+	if internaltypes.IsDefined(plan.EnableIdpAuthnSelection) {
```
henryrecker-pingidentity, line 66: No need for the IsDefined checks here - ValueBoolPointer already will return the correct values if one of these isn't defined (and they should be defined anyways since they both have a default value).

internal/resource/config/authenticationpolicies/settings/authentication_policies_settings_resource.go
```diff
@@ -0,0 +1,227 @@
...
+				Default:     booldefault.StaticBool(false),
+			},
+		},
+	}
+
+	id.ToSchema(&schema)
+	resp.Schema = schema
+}
+
+func addOptionalAuthenticationPoliciesSettingsFields(ctx context.Context, addRequest *client.AuthenticationPoliciesSettings, plan authenticationPoliciesSettingsModel) error {
```
henryrecker-pingidentity, line 64: No need to return an error from this method since it only will ever return nil. And `ctx` isn't used so it can be removed.

internal/resource/config/authenticationpolicies/settings/authentication_policies_settings_resource.go
```diff
@@ -0,0 +1,227 @@
...
+		return
+	}
+
+	createAuthenticationPoliciesSettings := client.NewAuthenticationPoliciesSettings()
+	err := addOptionalAuthenticationPoliciesSettingsFields(ctx, createAuthenticationPoliciesSettings, plan)
+	if err != nil {
+		resp.Diagnostics.AddError("Failed to add optional properties to add request for authentication policies settings", err.Error())
+		return
+	}
+	requestJson, err := createAuthenticationPoliciesSettings.MarshalJSON()
```
henryrecker-pingidentity, line 109: We decided to remove all these marshalJsons from the code so this should be removed along with any other marshalljson calls. If we're going to continue to scaffold resources with the code generator we need to update it to leave this out.

internal/resource/config/authenticationpolicies/settings/authentication_policies_settings_resource.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+	apiReadAuthenticationPoliciesSettings, httpResp, err := r.apiClient.AuthenticationPoliciesAPI.GetAuthenticationPolicySettings(config.ProviderBasicAuthContext(ctx, r.providerConfig)).Execute()
+
+	if err != nil {
+		if httpResp != nil && httpResp.StatusCode == 404 {
+			config.ReportHttpErrorAsWarning(ctx, &resp.Diagnostics, "An error occurred while getting the authentication policies settings", err, httpResp)
+			resp.State.RemoveResource(ctx)
+		} else {
+			config.ReportHttpError(ctx, &resp.Diagnostics, "An error occurred while getting the authentication policies settings", err, httpResp)
+		}
```
henryrecker-pingidentity, line 150: Need a return after this line. Otherwise it will try to read into state from an empty response after a 404

internal/resource/config/authenticationpolicies/settings/authentication_policies_settings_resource.go
```diff
@@ -0,0 +1,227 @@
...
+		resp.Diagnostics.AddError("Failed to add optional properties to add request for authentication policies settings", err.Error())
+		return
+	}
+	requestJson, err := createUpdateRequest.MarshalJSON()
+	if err == nil {
+		tflog.Debug(ctx, "Update request: "+string(requestJson))
+	}
+	updateAuthenticationPoliciesSettings = updateAuthenticationPoliciesSettings.Body(*createUpdateRequest)
+	updateAuthenticationPoliciesSettingsResponse, httpResp, err := r.apiClient.AuthenticationPoliciesAPI.UpdateAuthenticationPolicySettingsExecute(updateAuthenticationPoliciesSettings)
+	if err != nil && (httpResp == nil || httpResp.StatusCode != 404) {
```
henryrecker-pingidentity, line 197: Update doesn't need 404 logic like this

henryrecker-pingidentity (changes requested):

examples/resources/pingfederate_authentication_policies_settings/import.sh
PingDavidR: OK, I'll leave it here for consistency and we can do a global fix after researching this out.

internal/resource/config/authenticationpolicies/settings/authentication_policies_settings_common.go
PingDavidR: Done

internal/resource/config/authenticationpolicies/settings/authentication_policies_settings_resource.go
henryrecker-pingidentity: Still need to remove these

internal/resource/config/authenticationpolicies/settings/authentication_policies_settings_common.go
henryrecker-pingidentity: Still need to remove ctx from this method

internal/acctest/authenticationpoliciessettings/authentication_policies_settings_test.go
henryrecker-pingidentity: There should be four steps. Minimal -> Maximal -> Import -> back to Minimal.

internal/acctest/authenticationpoliciessettings/authentication_policies_settings_test.go
```diff
@@ -0,0 +1,112 @@
...
+// Attributes to test with. Add optional properties to test here if desired.
+type authenticationPoliciesSettingsResourceModel struct {
+	enableIdpAuthnSelection bool
+	enableSpAuthnSelection  bool
+}
+
+func TestAccAuthenticationPoliciesSettings(t *testing.T) {
+	resourceName := "myAuthenticationPoliciesSettings"
+	// No required resources, test empty resource
+	emptyResourceModel := authenticationPoliciesSettingsResourceModel{}
```
henryrecker-pingidentity, line 24: This won't result in actual empty HCL being sent, because you are still setting the attributes in the HCL you provide in the testAccAuthenticationPoliciesSettings method. You need to provide HCL with no attributes set. The log settings test does this with some optional HCL it inserts based on a boolean.

internal/acctest/authenticationpoliciessettings/authentication_policies_settings_test.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+data "pingfederate_authentication_policies_settings" "%[1]s" {
+  depends_on = [pingfederate_authentication_policies_settings.%[1]s]
+}`, resourceName,
+		optionalHcl,
+	)
+}
+
+// func testAccAuthenticationPoliciesSettings(resourceName string, resourceModel authenticationPoliciesSettingsResourceModel) string {
```
henryrecker-pingidentity, line 71: You can remove these comments

henryrecker-pingidentity (approved):

PR #156, opened 2024-01-04
Authentication api applications

wesleymccollam:


CHANGELOG.md
```diff
@@ -1,12 +1,17 @@
 # v0.5.0 (Unreleased)
+### BUG FIXES
+* **OAuth Client Resource:** Resolved issue where default issues on the `pingfederate_oauth_client` resource cause an invalid apply ([#](https://github.com/pingidentity/terraform-provider-pingfederate/pull/156))
```
henryrecker-pingidentity, line 3: Missing the number on the link text. Also is this supposed to say "default issues"?

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -256,13 +257,17 @@ func (r *oauthClientResource) Schema(ctx context.Context, req resource.SchemaReq
...
+				},
 			},
 			"restrict_scopes": schema.BoolAttribute{
 				Description: "Restricts this client's access to specific scopes.",
 				Computed:    true,
 				Optional:    true,
-				Default:     booldefault.StaticBool(false),
+				PlanModifiers: []planmodifier.Bool{
+					boolplanmodifier.UseStateForUnknown(),
+				},
```
henryrecker-pingidentity, line 270: What was the issue with these?

Makefile
```diff
@@ -51,15 +51,15 @@ endef
...
 # Set ACC_TEST_NAME to name of test in cli
 testoneacc:
-	$(call test_acc_env_vars) TF_ACC=1 go test ./internal/acctest/... -timeout 10m -run ${ACC_TEST_NAME} -v --count=1
+	$(call test_acc_env_vars) TF_ACC=1 go test ./internal/acctest/... -timeout 10m -run ${ACC_TEST_NAME} -v count=1
 
 testoneacccomplete: spincontainer testoneacc
 
 # Some tests can step on each other's toes so run those tests in single threaded mode. Run the rest in parallel
 testacc:
-	$(call test_acc_env_vars) TF_ACC=1 go test `go list ./internal/acctest/... | grep -v -e oauthauthserversettings -e oauthopenidconnectpolicy` -timeout 10m -v -p 4; \
```
henryrecker-pingidentity, line 60: I don't think we should always ignore caching with count=1 in this target. I get a lot of time save from caching.

internal/resource/config/authenticationapi/application/authentication_api_application_resource.go
```diff
@@ -0,0 +1,269 @@
...
+	updateAuthenticationApiApplication = updateAuthenticationApiApplication.Body(*createUpdateRequest)
+	updateAuthenticationApiApplicationResponse, httpResp, err := r.apiClient.AuthenticationApiAPI.UpdateApplicationExecute(updateAuthenticationApiApplication)
+	if err != nil {
+		config.ReportHttpError(ctx, &resp.Diagnostics, "An error occurred while updating an Authentication Api Application", err, httpResp)
+		return
+	}
+
+	// Read the response
+	var state authenticationApiApplicationModel
+	// This is a workaround for the fact that the API does not return the location of oauth client ClientForRedirectlessModeRef
```
henryrecker-pingidentity, line 233: I'd specify in this comment that it specifically applies to creates/updates, but location is returned normally on reads, which is why we are running an additional get here.

internal/resource/config/authenticationapi/application/authentication_api_application_resource.go
```diff
@@ -0,0 +1,269 @@
...
+func (r *authenticationApiApplicationResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+
+	var plan authenticationApiApplicationModel
+	diags := req.Plan.Get(ctx, &plan)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	updateAuthenticationApiApplication := r.apiClient.AuthenticationApiAPI.UpdateApplication(config.ProviderBasicAuthContext(ctx, r.providerConfig), plan.Id.ValueString())
```
henryrecker-pingidentity, line 216: I think we should generally be using the application_id attribute whenever possible, and tread id as just terraform information.

internal/resource/config/authenticationapi/application/authentication_api_application_resource.go
```diff
@@ -0,0 +1,269 @@
...
+
+	apiReadAuthenticationApiApplication, httpResp, err := r.apiClient.AuthenticationApiAPI.GetApplication(config.ProviderBasicAuthContext(ctx, r.providerConfig), state.ApplicationId.ValueString()).Execute()
+
+	if err != nil {
+		if httpResp != nil && httpResp.StatusCode == 404 {
+			config.ReportHttpErrorAsWarning(ctx, &resp.Diagnostics, "An error occurred while getting an Authentication Api Application", err, httpResp)
+			resp.State.RemoveResource(ctx)
+		} else {
+			config.ReportHttpError(ctx, &resp.Diagnostics, "An error occurred while getting an Authentication Api Application", err, httpResp)
+		}
```
henryrecker-pingidentity, line 195: Need a return after this line, otherwise it will continue despite these errors

internal/resource/config/authenticationapi/application/authentication_api_application_resource.go
```diff
@@ -0,0 +1,269 @@
...
+
+	providerCfg := req.ProviderData.(internaltypes.ResourceConfiguration)
+	r.providerConfig = providerCfg.ProviderConfig
+	r.apiClient = providerCfg.ApiClient
+
+}
+
+func addOptionalAuthenticationApiApplicationFields(ctx context.Context, addRequest *client.AuthnApiApplication, plan authenticationApiApplicationModel) error {
+
+	if internaltypes.IsDefined(plan.Description) {
```
henryrecker-pingidentity, line 115: Don't need isDefined checks for simple attributes like these (strings, ints, etc). You might not need it either for the set/list below either, though I'm not sure on that.

internal/resource/config/authenticationapi/application/authentication_api_application_resource.go
```diff
@@ -0,0 +1,269 @@
...
+		addRequest.Description = plan.Description.ValueStringPointer()
+	}
+
+	if internaltypes.IsDefined(plan.AdditionalAllowedOrigins) {
+		var slice []string
+		plan.AdditionalAllowedOrigins.ElementsAs(ctx, &slice, false)
+		addRequest.AdditionalAllowedOrigins = slice
+	}
+
+	if internaltypes.IsDefined(plan.ClientForRedirectlessModeRef) {
```
henryrecker-pingidentity, line 125: I think there should be a common method for this in the resourcelink package

henryrecker-pingidentity (changes requested):

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: These properties seemed to be always `false` in prior testing. When applying a different combination of attributes to make the correct ref in the authenticationApi/applications resource, this was not the case.

henryrecker-pingidentity (approved):

PR #169, opened 2024-01-18
remove default admin url suffix

wesleymccollam:


internal/types/definitions.go
```diff
@@ -8,6 +8,7 @@ import (
 // Configuration used by the provider and resources
 type ProviderConfiguration struct {
 	HttpsHost      string
+	AdminApiPath   string
```
henryrecker-pingidentity, line 11: I don't think we need this in this struct if it's never used in any resource

internal/provider/provider.go
```diff
@@ -312,7 +337,7 @@ func (p *pingfederateProvider) Configure(ctx context.Context, req provider.Confi
 	clientConfig.DefaultHeader["X-BypassExternalValidation"] = strconv.FormatBool(xBypassExternalValidation)
 	clientConfig.Servers = client.ServerConfigurations{
 		{
-			URL: httpsHost + "/pf-admin-api/v1",
+			URL: httpsHost + adminApiPath,
```
henryrecker-pingidentity, line 340: Do we need to have these as separate variables? Would it make more sense to just have one entire host variable including the admin api path?

henryrecker-pingidentity (approved):

internal/provider/provider.go
patrickcping: This is a breaking change for current users - I've asked Wes to revert

wesleymccollam: fixed and merged https://github.com/pingidentity/terraform-provider-pingfederate/pull/170

PR #182, opened 2024-02-01
Add extended properties resource

PingDavidR:
- closes PDI-1379


examples/resources/pingfederate_extended_properties/import.sh
```diff
@@ -0,0 +1,2 @@
+# "extendedPropertiesId" should be the ID of the extended properties to be imported
```
henryrecker-pingidentity, line 1: Since this appears to be edit-only, this comment should follow the format of other edit only import examples, such as server settings

examples/resources/pingfederate_extended_properties/resource.tf
```diff
@@ -0,0 +1,14 @@
+resource "pingfederate_extended_properties" "extendedPropertiesExample" {
+  items = [
+    {
+      name         = "extended_attribute"
+      description  = "attribute_1_description"
+      multi_valued = false
+    },
+    {
+      name         = "extended_attribute_3"
```
henryrecker-pingidentity, line 9: nit: Why is this 3 instead of 2?

internal/resource/config/extendedproperties/extended_properties_resource.go
```diff
@@ -0,0 +1,249 @@
...
+	Items types.List   `tfsdk:"items"`
+}
+
+// GetSchema defines the schema for the resource.
+func (r *extendedPropertiesResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	schema := schema.Schema{
+		Description: "Manages Extended Properties definitions",
+		Attributes: map[string]schema.Attribute{
+			"items": schema.ListNestedAttribute{
+				Description: "Extended Property definition that allows to store additional information about IdP/SP Connections and OAuth Clients.",
```
henryrecker-pingidentity, line 57: "Allows to store"

internal/resource/config/extendedproperties/extended_properties_resource.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	apiReadExtendedProperties, httpResp, err := r.apiClient.ExtendedPropertiesAPI.GetExtendedProperties(config.ProviderBasicAuthContext(ctx, r.providerConfig)).Execute()
+
+	if err != nil {
+		if httpResp != nil && httpResp.StatusCode == 404 {
+			config.ReportHttpErrorAsWarning(ctx, &resp.Diagnostics, "An error occurred while getting the extended properties", err, httpResp)
+			resp.State.RemoveResource(ctx)
+		} else {
+			config.ReportHttpError(ctx, &resp.Diagnostics, "An error occurred while getting the extended properties", err, httpResp)
+		}
```
henryrecker-pingidentity, line 185: Need a return after this line, otherwise it will continue despite the error

internal/acctest/config/extendedproperties/extended_properties_test.go
```diff
@@ -0,0 +1,129 @@
...
+		testClient := acctest.TestClient()
+		ctx := acctest.TestBasicAuthContext()
+		response, _, err := testClient.ExtendedPropertiesAPI.GetExtendedProperties(ctx).Execute()
+
+		if err != nil {
+			return err
+		}
+
+		// Verify that attributes have expected values
+		if len(response.GetItems()) > 0 {
```
henryrecker-pingidentity, line 107: You could put all of this in one if statement rather than splitting into three.

It would also be good to have a check that the items length is zero as expected when passing in the initial resource model

examples/resources/pingfederate_extended_properties/import.sh
```diff
@@ -1,2 +1,2 @@
-# "extendedPropertiesId" should be the ID of the extended properties to be imported
+# This resource is singleton, so the value of "id" doesn't matter - it is just a placeholder, and required by Terraform
 terraform import pingfederate_extended_properties.extendedProperties extendedPropertiesId
```
henryrecker-pingidentity, line 2: The comment describes "id" but "extendedPropertiesId" is used on this line

internal/acctest/config/extendedproperties/extended_properties_test.go
```diff
@@ -103,22 +103,26 @@ func testAccCheckExpectedExtendedPropertiesAttributes(extendedProperty *client.E
...
 
+		// Check for no items if empty object sent
+		if extendedProperty.Name == nil {
+			if len(response.GetItems()) > 0 {
+				return fmt.Errorf("Empty items object sent, expected no items to be returned, but got %d", len(response.GetItems()))
+			}
+			return nil
+		}
+
 		// Verify that attributes have expected values
```
henryrecker-pingidentity, line 114: This probably won't matter, but this could potentially miss an error. If we expect more than one item returned, this won't error out if none are returned. Because the checks only happen inside the if statement.

henryrecker-pingidentity (approved):

PR #183, opened 2024-02-02
OAuth2 client credentials, access token support, pingfederate_oauth_client resource bug fix

wesleymccollam:
- OAuth2 Client Credentials Flow Support
- Access Token Support
- `pingfederate_oauth_client` resource `grant_types` bug fix


internal/acctest/acctest.go
```diff
@@ -59,22 +66,28 @@ func TestClient() *client.APIClient {
...
+}
+
 func TestBasicAuthContext() context.Context {
 	ctx := context.Background()
 	return config.BasicAuthContext(ctx, os.Getenv("PINGFEDERATE_PROVIDER_USERNAME"), os.Getenv("PINGFEDERATE_PROVIDER_PASSWORD"))
 }
 
+func TestOauth2Context() context.Context {
+	ctx := context.Background()
+	return config.OAuthContext(ctx, GetTransport(), os.Getenv("PINGFEDERATE_PROVIDER_OAUTH_TOKEN_URL"), os.Getenv("PINGFEDERATE_PROVIDER_OAUTH_CLIENT_ID"), os.Getenv("PINGFEDERATE_PROVIDER_OAUTH_CLIENT_SECRET"), []string{os.Getenv("PINGFEDERATE_PROVIDER_OAUTH_SCOPES")})
```
henryrecker-pingidentity, line 88: It'd be good to have some error reporting here if any of these env vars aren't set.

internal/acctest/authentication/access_token_test.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+type Response struct {
+	AccessToken string `json:"access_token"`
+}
+
+func getAccessToken() {
+	oauthClientId := os.Getenv("PINGFEDERATE_PROVIDER_OAUTH_CLIENT_ID")
+	oauthClientSecret := os.Getenv("PINGFEDERATE_PROVIDER_OAUTH_CLIENT_SECRET")
+	os.Getenv("PINGFEDERATE_PROVIDER_OAUTH_TOKEN_URL")
```
henryrecker-pingidentity, line 27: What's the purpose of these two calls?

internal/acctest/authentication/access_token_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/pingidentity/terraform-provider-pingfederate/internal/provider"
+)
+
+type Response struct {
+	AccessToken string `json:"access_token"`
+}
+
+func getAccessToken() {
+	oauthClientId := os.Getenv("PINGFEDERATE_PROVIDER_OAUTH_CLIENT_ID")
+	oauthClientSecret := os.Getenv("PINGFEDERATE_PROVIDER_OAUTH_CLIENT_SECRET")
```
henryrecker-pingidentity, line 26: I'd add some error checking to confirm these variables are set

internal/acctest/authentication/access_token_test.go
```diff
@@ -0,0 +1,101 @@
...
+	client := &http.Client{Transport: acctest.GetTransport()}
+	clientInfo := fmt.Sprintf("client_id=%s&grant_type=client_credentials&client_secret=%s&scope=email", oauthClientId, oauthClientSecret)
+	jsonBodyReader := strings.NewReader(clientInfo)
+	resp, err := client.Post(tokenRequestUrl, "application/x-www-form-urlencoded", jsonBodyReader)
+	if err != nil {
+		return
+	}
+
+	os.Unsetenv("PINGFEDERATE_PROVIDER_OAUTH_CLIENT_ID")
+	os.Unsetenv("PINGFEDERATE_PROVIDER_OAUTH_CLIENT_SECRET")
```
henryrecker-pingidentity, line 42: Do we still need these with the way the tests are split up in the Makefile?

internal/acctest/authentication/access_token_test.go
```diff
@@ -0,0 +1,101 @@
...
+	//#nosec G101
+	tokenRequestUrl := "https://localhost:9031/as/token.oauth2"
+
+	//#nosec G402
+	client := &http.Client{Transport: acctest.GetTransport()}
+	clientInfo := fmt.Sprintf("client_id=%s&grant_type=client_credentials&client_secret=%s&scope=email", oauthClientId, oauthClientSecret)
+	jsonBodyReader := strings.NewReader(clientInfo)
+	resp, err := client.Post(tokenRequestUrl, "application/x-www-form-urlencoded", jsonBodyReader)
+	if err != nil {
+		return
```
henryrecker-pingidentity, line 38: This should propagate the error up so that there is useful output when it fails. Same for the other returns in this method

internal/acctest/authentication/access_token_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+func testAccATVirtualHostNames(resourceName string) string {
+	return fmt.Sprintf(`
+resource "pingfederate_virtual_host_names" "%[1]s" {
+  virtual_host_names = %[2]s
+}
+data "pingfederate_virtual_host_names" "%[1]s" {
+  depends_on = [pingfederate_virtual_host_names.%[1]s]
+}`, resourceName,
+		acctest.StringSliceToTerraformString([]string{"test"}),
```
henryrecker-pingidentity, line 83: If this is hardcoded you can just put it directly in the string parameter rather than formatting it in

internal/acctest/authentication/access_token_test.go
```diff
@@ -0,0 +1,101 @@
...
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() { acctest.ConfigurationPreCheck(t) },
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
+			"pingfederate": providerserver.NewProtocol6WithError(provider.NewTestProvider()),
+		},
+		Steps: []resource.TestStep{
+			{
+				PreConfig: getAccessToken,
+				Config:    testAccATVirtualHostNames("virtualHostNames"),
+				Check:     testAccATGetVirtualHostNames(),
```
henryrecker-pingidentity, line 69: We don't really need a check in this test. We just want to be sure the Config step succeeds without a terraform error.

internal/acctest/authentication/oauth_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework/providerserver"
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/pingidentity/terraform-provider-pingfederate/internal/acctest"
+	"github.com/pingidentity/terraform-provider-pingfederate/internal/provider"
+)
+
+func getOAuthEnvVars() {
```
henryrecker-pingidentity, line 16: This method doesn't appear to do anything, it just retrieves the vars but doesn't save them anywhere

internal/acctest/authentication/oauth_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+func testAccOAuthVirtualHostNames(resourceName string) string {
+	return fmt.Sprintf(`
+resource "pingfederate_virtual_host_names" "%[1]s" {
+  virtual_host_names = %[2]s
+}
+data "pingfederate_virtual_host_names" "%[1]s" {
+  depends_on = [pingfederate_virtual_host_names.%[1]s]
+}`, resourceName,
+		acctest.StringSliceToTerraformString([]string{"test"}),
```
henryrecker-pingidentity, line 47: Same comment as the other test here, just put this in the string itself. In fact there's not really a need to have resourceName formatted in either, you can just hardcode the entire HCL.

internal/acctest/authentication/oauth_test.go
```diff
@@ -0,0 +1,64 @@
...
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() { acctest.ConfigurationPreCheck(t) },
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
+			"pingfederate": providerserver.NewProtocol6WithError(provider.NewTestProvider()),
+		},
+		Steps: []resource.TestStep{
+			{
+				PreConfig: getOAuthEnvVars,
+				Config:    testAccOAuthVirtualHostNames("virtualHostNames"),
+				Check:     testAccOAuthGetVirtualHostNames(),
```
henryrecker-pingidentity, line 33: No need for a check here either

internal/acctest/config/oauth/openidconnect/policy/oauth_open_id_connect_policy_test.go
```diff
@@ -33,6 +35,33 @@ type oauthOpenIdConnectPoliciesResourceModel struct {
 	reissueIdTokenInHybridFlow  *bool
 }
 
+// This is due to a bug in PingFederate that doesn't allow the OAuth client to set "None" as the OIDC Policy
+func deleteOauthClient() {
+	//#nosec G402
+	client := &http.Client{Transport: acctest.GetTransport()}
```
henryrecker-pingidentity, line 41: You could use our Go client to make this a bit easier

internal/acctest/config/oauth/openidconnect/policy/oauth_open_id_connect_policy_test.go
```diff
@@ -33,6 +35,33 @@ type oauthOpenIdConnectPoliciesResourceModel struct {
 	reissueIdTokenInHybridFlow  *bool
 }
 
+// This is due to a bug in PingFederate that doesn't allow the OAuth client to set "None" as the OIDC Policy
```
henryrecker-pingidentity, line 38: Could you provide a little more detail in this comment? The actual issue comes up when you try to delete the oidc policy, right?

internal/provider/provider.go
```diff
@@ -117,14 +122,53 @@ func (p *pingfederateProvider) Schema(_ context.Context, _ provider.SchemaReques
 				MarkdownDescription: "Path for PingFederate Admin API. Default value can be set with the `PINGFEDERATE_PROVIDER_ADMIN_API_PATH` environment variable. If no value is supplied, the value used will be `/pf-admin-api/v1`.",
 				Optional:            true,
 			},
+			"access_token": schema.StringAttribute{
```
henryrecker-pingidentity, line 125: Why have this separate from the oauth nested attribute? We probably should compare with pingone and/or other providers like Scott mentioned to follow a typical pattern.

Maybe username and password should be in a "basicauth" nested attribute? (though that would be breaking so it might be something that has to wait).

internal/provider/provider.go
```diff
@@ -117,14 +122,53 @@ func (p *pingfederateProvider) Schema(_ context.Context, _ provider.SchemaReques
...
+					stringvalidator.ConflictsWith(path.MatchRoot("oauth")),
+					stringvalidator.ConflictsWith(path.MatchRoot("username")),
+					stringvalidator.ConflictsWith(path.MatchRoot("password")),
+				},
+			},
 			"username": schema.StringAttribute{
 				MarkdownDescription: "Username for PingFederate Admin user. Default value can be set with the `PINGFEDERATE_PROVIDER_USERNAME` environment variable.",
 				Optional:            true,
+				Validators: []validator.String{
+					stringvalidator.ConflictsWith(path.MatchRoot("oauth")),
```
henryrecker-pingidentity, line 139: For these conflicts, it would be good to describe the conflict cases in the attribute description, since the errors terraform outputs on these tend to be a little cryptic

internal/provider/provider.go
```diff
@@ -117,14 +122,53 @@ func (p *pingfederateProvider) Schema(_ context.Context, _ provider.SchemaReques
...
 			"password": schema.StringAttribute{
 				MarkdownDescription: "Password for PingFederate Admin user. Default value can be set with the `PINGFEDERATE_PROVIDER_PASSWORD` environment variable.",
 				Optional:            true,
 				Sensitive:           true,
+				Validators: []validator.String{
+					stringvalidator.ConflictsWith(path.MatchRoot("oauth")),
+				},
+			},
+			"oauth": schema.SingleNestedAttribute{
+				MarkdownDescription: "OAuth configuration for requesting access token. Default values can be set with the `PINGFEDERATE_PROVIDER_OAUTH_*` environment variables.",
```
henryrecker-pingidentity, line 151: Is there a name for the oauth flow we are using here, and if so can we put it in the description? Might be a dumb question.

internal/provider/provider.go
```diff
@@ -117,14 +122,53 @@ func (p *pingfederateProvider) Schema(_ context.Context, _ provider.SchemaReques
...
+				Optional:            true,
+				Attributes: map[string]schema.Attribute{
+					"client_id": schema.StringAttribute{
+						MarkdownDescription: "OAuth client ID for requesting access token. Default value can be set with the `PINGFEDERATE_PROVIDER_OAUTH_CLIENT_ID` environment variable.",
+						Required:            true,
+					},
+					"client_secret": schema.StringAttribute{
+						MarkdownDescription: "OAuth client secret for requesting access token. Default value can be set with the `PINGFEDERATE_PROVIDER_OAUTH_CLIENT_SECRET` environment variable.",
+						Required:            true,
```
henryrecker-pingidentity, line 160: This should be sensitive. Not sure if the other ones might benefit from being sensitive as well.

internal/resource/config/pingoneconnection/ping_one_connection_resource.go
```diff
@@ -0,0 +1,295 @@
+package pingoneconnection
```
henryrecker-pingidentity, line 1: Is this file supposed to be here?

internal/provider/provider.go
```diff
@@ -196,50 +240,147 @@ func (p *pingfederateProvider) Configure(ctx context.Context, req provider.Confi
...
 	} else {
-		if config.Username.IsNull() {
-			username = os.Getenv("PINGFEDERATE_PROVIDER_USERNAME")
+		if config.AccessToken.IsNull() {
+			accessToken = os.Getenv("PINGFEDERATE_PROVIDER_ACCESS_TOKEN")
 		} else {
-			username = config.Username.ValueString()
+			accessToken = config.AccessToken.ValueString()
 		}
-		if username == "" {
```
henryrecker-pingidentity, line 213: It might be useful to still have some log message when any one of these vars is empty

internal/provider/provider.go
```diff
@@ -196,50 +240,147 @@ func (p *pingfederateProvider) Configure(ctx context.Context, req provider.Confi
...
+	var password string
+	if config.Password.IsNull() {
+		password = os.Getenv("PINGFEDERATE_PROVIDER_PASSWORD")
+	} else {
+		password = config.Password.ValueString()
+	}
+
+	// Check if the user has provided an access token to the provider
+	var accessToken string
+	if config.AccessToken.IsUnknown() {
```
henryrecker-pingidentity, line 261: Why is the Unknown check removed for username & password?

internal/provider/provider.go
```diff
@@ -196,50 +240,147 @@ func (p *pingfederateProvider) Configure(ctx context.Context, req provider.Confi
...
+	}
+
+	// Check if the user has provided an OAuth configuration to the provider
+	var hasOauthConfig bool = false
+	var clientId string
+	var clientSecret string
+	var scopes []string
+	var tokenUrl string
+	if username == "" && password == "" && accessToken == "" {
+		configClientId, ok := config.OAuth.Attributes()["client_id"].(types.String)
```
henryrecker-pingidentity, line 282: What happens here if the user doesn't set the oauth block at all (if they haven't set any authentication attributes)? It seems like this could end up trying to cast a nil and panicking.

internal/provider/provider.go
```diff
@@ -196,50 +240,147 @@ func (p *pingfederateProvider) Configure(ctx context.Context, req provider.Confi
...
+		if ok {
+			clientId = configClientId.ValueString()
+		} else {
+			clientId = os.Getenv("PINGFEDERATE_PROVIDER_OAUTH_CLIENT_ID")
+			if clientId == "" {
+				resp.Diagnostics.AddError(
+					"Unable to find client_id",
+					"client_id cannot be an empty string. Either set it in the configuration or use the PINGFEDERATE_PROVIDER_OAUTH_CLIENT_ID environment variable.",
```
henryrecker-pingidentity, line 290: These error messages seem not quite accurate, since at this point the user might not have set any authentication-related attributes at all. They may not specifically be wanting to use oauth. So we would want to indicate that they need to either configure basic auth, provide an access token, or configure oauth.

internal/provider/provider.go
```diff
@@ -196,50 +240,147 @@ func (p *pingfederateProvider) Configure(ctx context.Context, req provider.Confi
...
+				)
+			}
+		}
+
+		configClientSecret, ok := config.OAuth.Attributes()["client_secret"].(types.String)
+		if ok {
+			clientSecret = configClientSecret.ValueString()
+		} else {
+			clientSecret = os.Getenv("PINGFEDERATE_PROVIDER_OAUTH_CLIENT_SECRET")
+			if clientSecret == "" {
```
henryrecker-pingidentity, line 300: Should these checks be outside of the else block? Since they shouldn't be allowed to set an empty string in the provider config or via env vars.

internal/provider/provider.go
```diff
@@ -196,50 +240,147 @@ func (p *pingfederateProvider) Configure(ctx context.Context, req provider.Confi
...
-		if config.Password.IsNull() {
-			password = os.Getenv("PINGFEDERATE_PROVIDER_PASSWORD")
-		} else {
-			password = config.Password.ValueString()
+		// If user has not provided an OAuth configuration or access token, they must provide username and password
+		if !hasOauthConfig && accessToken == "" {
+			if username == "" {
+				resp.Diagnostics.AddError(
+					"Unable to find username",
+					"username cannot be an empty string. Either set it in the configuration or use the PINGFEDERATE_PROVIDER_USERNAME environment variable.",
```
henryrecker-pingidentity, line 352: I might just combine a lot of this validation into one error message. If anything mutually exclusive is set at the same time, add one common error message that the user must choose between basic auth, access token, and oauth, along with the corresponding env vars to do so.

internal/provider/provider.go
```diff
@@ -360,9 +520,8 @@ func (p *pingfederateProvider) Configure(ctx context.Context, req provider.Confi
 	clientConfig.UserAgentSuffix = pointers.String(userAgentSuffix)
 	resourceConfig.ApiClient = client.NewAPIClient(clientConfig)
 	resp.ResourceData = resourceConfig
-	resp.DataSourceData = resourceConfig
-
 	tflog.Info(ctx, "Configured PingFederate client", map[string]interface{}{"success": true})
+	resp.DataSourceData = resourceConfig
```
henryrecker-pingidentity, line 524: Any reason this was moved down?

internal/resource/config/utils.go
```diff
@@ -22,7 +22,44 @@ func BasicAuthContext(ctx context.Context, username, password string) context.Co
...
+		Scopes:       scopes,
+	})
+}
+
+// Get an OAuth context from a ProviderConfiguration
+func ProviderOAuthContext(ctx context.Context, providerConfig internaltypes.ProviderConfiguration) context.Context {
+	return OAuthContext(ctx, providerConfig.Transport, *providerConfig.TokenUrl, *providerConfig.ClientId, *providerConfig.ClientSecret, providerConfig.Scopes)
+}
+
+func DetermineAuthContext(ctx context.Context, providerConfig internaltypes.ProviderConfiguration) context.Context {
```
henryrecker-pingidentity, line 54: I'd probably just name this AuthContext

internal/types/definitions.go
```diff
@@ -1,15 +1,23 @@
...
 
 import (
+	"net/http"
+
 	client "github.com/pingidentity/pingfederate-go-client/v1200/configurationapi"
 	"github.com/pingidentity/terraform-provider-pingfederate/internal/version"
 )
 
 // Configuration used by the provider and resources
 type ProviderConfiguration struct {
```
henryrecker-pingidentity, line 11: Maybe we could have this struct just hold the contexts themselves, rather than the vars to create the contexts? Not 100% on whether there would be any issues with that.

Makefile
```diff
@@ -45,27 +45,44 @@ removetestcontainer:
...
+	PINGFEDERATE_PROVIDER_OAUTH_CLIENT_ID=test PINGFEDERATE_PROVIDER_OAUTH_CLIENT_SECRET=2FederateM0re! PINGFEDERATE_PROVIDER_OAUTH_TOKEN_URL=https://localhost:9031/as/token.oauth2 PINGFEDERATE_PROVIDER_OAUTH_SCOPES=email
 endef
 
 # Set ACC_TEST_NAME to name of test in cli
 testoneacc:
-	$(call test_acc_env_vars) TF_ACC=1 go test ./internal/acctest/config/... -timeout 10m -run ${ACC_TEST_NAME} -v count=1
+	$(call test_acc_common_env_vars) $(call test_acc_basic_auth_env_vars) TF_ACC=1 go test ./internal/acctest/... -timeout 10m -run ${ACC_TEST_NAME} -v count=1
 
-testoneacccomplete: spincontainer testoneacc
+testoneacccomplete: spincontainer testauthacc testacc
```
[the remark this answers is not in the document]
wesleymccollam: whoops!

internal/acctest/authentication/access_token_test.go
wesleymccollam: looks like we don't. I'll remove them

internal/acctest/authentication/access_token_test.go
wesleymccollam: leaving this for sanity checking the client side of things

internal/acctest/authentication/oauth_test.go
wesleymccollam: leaving this in for client testing

internal/provider/provider.go
wesleymccollam: we can separate all auths in `v1.0.0`

internal/provider/provider.go
wesleymccollam: nice catch!

internal/resource/config/pingoneconnection/ping_one_connection_resource.go
wesleymccollam: oops, left from another branch in the works, good catch.

internal/provider/provider.go
wesleymccollam: not sure, adding this back

internal/provider/provider.go
wesleymccollam: the logic for this will be update to not include checks for the other properties, but rather validated later

internal/provider/provider.go
wesleymccollam: good catch

internal/provider/provider.go
wesleymccollam: zero. I was doing a multi-line edit and this got moved and I didn't realize it was out of place.. will move this back

internal/types/definitions.go
wesleymccollam: I'll add a GA case for this.

internal/acctest/authentication/access_token_test.go
```diff
@@ -0,0 +1,106 @@
...
+func getAccessToken(t *testing.T) {
+	//#nosec G101
+	tokenRequestUrl := "https://localhost:9031/as/token.oauth2"
+
+	//#nosec G402
+	client := &http.Client{Transport: acctest.GetTransport()}
+	envVars, errors := authentication.TestEnvVarSlice([]string{"PINGFEDERATE_PROVIDER_OAUTH_CLIENT_ID", "PINGFEDERATE_PROVIDER_OAUTH_CLIENT_SECRET"}, "access_token_test.go")
+	if len(errors) > 0 {
+		for _, err := range errors {
+			fmt.Println(err)
```
henryrecker-pingidentity, line 34: You can use the error functions on t rather than printing directly here

internal/acctest/authentication/access_token_test.go
henryrecker-pingidentity: I still see these in the method

internal/acctest/authentication/test_env_var_check.go
```diff
@@ -0,0 +1,19 @@
+package authentication
+
+import (
+	"errors"
+	"os"
+)
+
+func TestEnvVarSlice(vars []string, fileName string) (map[string]string, []error) {
```
henryrecker-pingidentity, line 8: I'd pass in the testing object here so you can FailNow when there are errors.

henryrecker-pingidentity:
It might be nice to have a test that confirms that the provider gracefully fails when no auth is configured at all, with the expected error message.

internal/provider/provider.go
```diff
@@ -233,13 +285,133 @@ func (p *pingfederateProvider) Configure(ctx context.Context, req provider.Confi
...
+			clientSecret = configClientSecret.ValueString()
+		} else {
+			clientSecret = os.Getenv("PINGFEDERATE_PROVIDER_OAUTH_CLIENT_SECRET")
+		}
+
+		configScopes, ok := config.OAuth.Attributes()["scopes"].(types.List)
+		if ok {
+			configScopes.ElementsAs(ctx, &scopes, false)
+		} else {
+			scopes = []string{os.Getenv("PINGFEDERATE_PROVIDER_OAUTH_SCOPES")}
```
henryrecker-pingidentity, line 344: This doesn't allow for users to define multiple scopes - they can only set at most one scope

internal/provider/provider.go
```diff
@@ -233,13 +285,133 @@ func (p *pingfederateProvider) Configure(ctx context.Context, req provider.Confi
...
+				returnAuthAttributeDiagsError("username", "basic", "PINGFEDERATE_PROVIDER_USERNAME", resp)
+			}
+
+			if password == "" {
+				returnAuthAttributeDiagsError("password", "basic", "PINGFEDERATE_PROVIDER_PASSWORD", resp)
+			}
+		}
+
+		// If user has not provided username and password or an OAuth configuration, they must provide an access token
+		if hasAccessTokenAuth && (!hasBasicAuth || !hasOauthConfig) {
```
henryrecker-pingidentity, line 389: What if the user has set username, password, and an auth token? Looks like this will accept that and end up setting all three on the provider config.

internal/provider/provider.go
```diff
@@ -233,13 +285,133 @@ func (p *pingfederateProvider) Configure(ctx context.Context, req provider.Confi
...
+			hasOauthConfig = false
+		}
+	}
+
+	// Validate the configuration
+	// User must provide a username and password, access token, or OAuth configuration to the provider
+	if !hasAccessTokenAuth && !hasBasicAuth && !hasOauthConfig {
+		resp.Diagnostics.AddError(
+			"Unable to find username and password, access_token, or OAuth authentication configuration",
+			"The username and password, access_token, or oauth configuration required values were not supplied. Either set them in the configuration or use the PINGFEDERATE_PROVIDER_* environment variables.",
```
henryrecker-pingidentity, line 369: I think we are losing some information here. Say they are using oauth but forgot to set tokenUrl. It'd be good if our message says exactly that, but they'd get this more generic message.

internal/provider/provider.go
```diff
@@ -233,13 +285,133 @@ func (p *pingfederateProvider) Configure(ctx context.Context, req provider.Confi
...
+
+		// If user has not provided username and password or an OAuth configuration, they must provide an access token
+		if hasAccessTokenAuth && (!hasBasicAuth || !hasOauthConfig) {
+			if accessToken == "" {
+				returnAuthAttributeDiagsError("access_token", "access token", "PINGFEDERATE_PROVIDER_ACCESS_TOKEN", resp)
+			}
+		}
+
+		// If user has not provided username and password or an access token, they must provide an OAuth configuration
+		if (!hasBasicAuth || !hasAccessTokenAuth) && hasOauthConfig {
```
henryrecker-pingidentity, line 396: Similar issue here - what if they provided username, password, and full oauth config?

internal/provider/provider.go
```diff
@@ -233,13 +285,133 @@ func (p *pingfederateProvider) Configure(ctx context.Context, req provider.Confi
...
+			if clientId == "" {
+				returnAuthAttributeDiagsError("client_id", "OAuth", "PINGFEDERATE_PROVIDER_OAUTH_CLIENT_ID", resp)
+			}
+			if clientSecret == "" {
+				returnAuthAttributeDiagsError("client_secret", "OAuth", "PINGFEDERATE_PROVIDER_OAUTH_CLIENT_SECRET", resp)
+			}
+			if tokenUrl == "" {
+				returnAuthAttributeDiagsError("token_url", "OAuth", "PINGFEDEATE_PROVIDER_OAUTH_TOKEN_URL", resp)
+			}
+		} else if hasOauthConfig && (!hasBasicAuth || !hasAccessTokenAuth) {
```
henryrecker-pingidentity, line 406: This else looks impossible to reach, because it has the same condition as the if

internal/provider/provider.go
```diff
@@ -117,36 +126,117 @@ func (p *pingfederateProvider) Schema(_ context.Context, _ provider.SchemaReques
...
+					stringvalidator.AlsoRequires(path.MatchRoot("token_url")),
+				},
+			},
+			"client_secret": schema.StringAttribute{
+				MarkdownDescription: "OAuth client secret for requesting access token. Default value can be set with the `PINGFEDERATE_PROVIDER_OAUTH_CLIENT_SECRET` environment variable.",
 				Optional:            true,
 				Sensitive:           true,
+				Validators: []validator.String{
+					stringvalidator.ConflictsWith(path.MatchRoot("username")),
+					stringvalidator.ConflictsWith(path.MatchRoot("password")),
```
henryrecker-pingidentity, line 163: Should conflict with access token

internal/provider/provider.go
```diff
@@ -233,13 +324,182 @@ func (p *pingfederateProvider) Configure(ctx context.Context, req provider.Confi
...
+			"Unable to connect to the PingFederate Server",
+			"Cannot use unknown value as scopes",
+		)
+	} else {
+		if config.Scopes.IsNull() {
+			scopes = strings.Split(os.Getenv("PINGFEDERATE_PROVIDER_OAUTH_SCOPES"), ",")
+		} else {
+			config.Scopes.ElementsAs(ctx, &scopes, false)
+		}
+		if len(scopes) <= 1 {
```
henryrecker-pingidentity, line 412: What if the user intentionally wants to set a single scope?

examples/doc-examples/access-token.tf
```diff
@@ -0,0 +1,8 @@
+provider "pingfederate" {
+  access_token                        = "accesstokenvaluefromclient"
+  https_host                          = "https://localhost:9999"
+  admin_api_path                      = "/pf-admin-api/v1"
+  insecure_trust_all_tls              = true
```
henryrecker-pingidentity, line 5: These new example files are great. It might be good to remove the `insecure_trust_all_tls` and `x_bypass_external_validation_header` from these since they're more for development and not relevant to authentication.

henryrecker-pingidentity (approved):
Looks good, just the one comment about the new examples.

PR #189, opened 2024-02-12
change to local server profiles from github repo

wesleymccollam:
container deployment method changed to repo server profiles


server-profiles/11.2.2/instance/bin/run.properties.subst
```diff
@@ -0,0 +1,561 @@
+# PingFederate Initialization Properties File
```
henryrecker-pingidentity, line 1: Do we need these entire properties files? Could we remove them, or just set a few necessary properties via environment variables?

server-profiles/11.2.2/motd
```diff
@@ -0,0 +1,11 @@
+
```
henryrecker-pingidentity, line 1: I'd remove non-functional files from the profiles like this one, we don't need this info in our provider repo.

server-profiles/11.3.0/README.md
```diff
@@ -0,0 +1,8 @@
+# Purpose
```
henryrecker-pingidentity, line 1: What's the difference between 11.3.0 and 11.3.4 profiles?

wesleymccollam: 11.2.2 and 11.2.7 are the same and are the only versions that have a different data.json.subst file. The multiple version folders are mainly just to simplify everything using the existing `PINGFEDERATE_PROVIDER_PRODUCT_VERSION` environment variable.

henryrecker-pingidentity: I'd like to avoid the fully duplicated profiles. Possibly with a separate variable for the profile version, or maybe it's possible by mounting individual files

wesleymccollam: agreed

server-profiles/11.2.2/instance/bin/run.properties.subst
wesleymccollam: added some env var substitution for the ones available. This is now in one file too, so not much maintenance required and it shouldn't change often (or at all)

server-profiles/11.2.2/motd
wesleymccollam: removed these

server-profiles/11.3.0/README.md
wesleymccollam: added shared profile with individual version data.json.subst files

henryrecker-pingidentity (approved):

PR #190, opened 2024-02-13
incomingProxySettings resource

PingDavidR:
- closes PDI-1381


internal/resource/config/incomingproxysettings/incoming_proxy_settings_resource.go
```diff
@@ -0,0 +1,285 @@
...
+				Description: "While the proxy server is configured to pass client certificates as HTTP request headers, specify the header name here.",
+				Computed:    false,
+				Optional:    true,
+			},
+			"client_cert_chain_ssl_header_name": schema.StringAttribute{
+				Description: "While the proxy server is configured to pass client certificates as HTTP request headers, specify the chain header name here.",
+				Computed:    false,
+				Optional:    true,
+			},
+			"proxy_terminates_https_conns": schema.BoolAttribute{
```
henryrecker-pingidentity, line 89: We should set a default here in the schema if it is computed and optional

internal/resource/config/incomingproxysettings/incoming_proxy_settings_resource.go
```diff
@@ -0,0 +1,285 @@
...
+		},
+	}
+
+	id.ToSchema(&schema)
+	resp.Schema = schema
+}
+
+func addOptionalIncomingProxySettingsFields(ctx context.Context, addRequest *client.IncomingProxySettings, plan incomingProxySettingsResourceModel) error {
+
+	if internaltypes.IsDefined(plan.ForwardedIpAddressHeaderName) {
```
henryrecker-pingidentity, line 103: Since these are all primitive values, you don't need any of the IsDefined checks in this method.

You can also update this method to not return error, since it never returns anything but nil.

internal/resource/config/incomingproxysettings/incoming_proxy_settings_resource.go
```diff
@@ -0,0 +1,285 @@
...
+		return
+	}
+
+	providerCfg := req.ProviderData.(internaltypes.ResourceConfiguration)
+	r.providerConfig = providerCfg.ProviderConfig
+	r.apiClient = providerCfg.ApiClient
+
+}
+
+func readIncomingProxySettingsResponse(ctx context.Context, r *client.IncomingProxySettings, state *incomingProxySettingsResourceModel, existingId *string) diag.Diagnostics {
```
henryrecker-pingidentity, line 151: This method doesn't need to return diagnostics since it never adds any diagnostic values

henryrecker-pingidentity (changes requested):

internal/resource/config/incomingproxysettings/incoming_proxy_settings_resource.go
```diff
@@ -0,0 +1,279 @@
...
+
+	id.ToSchema(&schema)
+	resp.Schema = schema
+}
+
+func addOptionalIncomingProxySettingsFields(ctx context.Context, addRequest *client.IncomingProxySettings, plan incomingProxySettingsResourceModel) {
+
+	addRequest.ForwardedIpAddressHeaderName = plan.ForwardedIpAddressHeaderName.ValueStringPointer()
+
+	if internaltypes.IsDefined(plan.ForwardedIpAddressHeaderIndex) {
```
henryrecker-pingidentity, line 108: These IsDefined checks aren't needed

internal/resource/config/incomingproxysettings/incoming_proxy_settings_resource.go
```diff
@@ -0,0 +1,279 @@
...
+
+	providerCfg := req.ProviderData.(internaltypes.ResourceConfiguration)
+	r.providerConfig = providerCfg.ProviderConfig
+	r.apiClient = providerCfg.ApiClient
+
+}
+
+func (r *incomingProxySettingsResource) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
+	var plan *incomingProxySettingsResourceModel
+	req.Plan.Get(ctx, &plan)
```
henryrecker-pingidentity, line 142: You can record the diagnostics of the Plan.Get call, with something like `resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)`

internal/resource/config/incomingproxysettings/incoming_proxy_settings_resource.go
```diff
@@ -0,0 +1,292 @@
...
+	providerCfg := req.ProviderData.(internaltypes.ResourceConfiguration)
+	r.providerConfig = providerCfg.ProviderConfig
+	r.apiClient = providerCfg.ApiClient
+
+}
+
+func (r *incomingProxySettingsResource) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
+	var plan *incomingProxySettingsResourceModel
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	req.Plan.Get(ctx, &plan)
```
henryrecker-pingidentity, line 140: You're getting the plan twice here

henryrecker-pingidentity (approved):

PR #199, opened 2024-02-28
authentication selector resource

wesleymccollam:


CHANGELOG.md
```diff
@@ -3,9 +3,10 @@
 * `location` property in `resourceLink` object types for all Resources and DataSources. This property will be removed in a future release. ([#195](https://github.com/pingidentity/terraform-provider-pingfederate/pull/195))
 
 ### Resources
+* **New Resource:** `pingfederate_authentication_selector ` ([#198](https://github.com/pingidentity/terraform-provider-pingfederate/pull/198))
```
henryrecker-pingidentity, line 6: This PR is 199

internal/resource/config/authenticationselector/authentication_selector_resource.go
```diff
@@ -0,0 +1,353 @@
...
+					objectplanmodifier.RequiresReplace(),
+				},
+			},
+			"parent_ref": schema.SingleNestedAttribute{
+				Description: "The reference to this plugin's parent instance. The parent reference is only accepted if the plugin type supports parent instances.",
+				Optional:    true,
+				Attributes:  resourcelink.ToSchemaNoLocation(),
+			},
+			"configuration": pluginconfiguration.ToSchema(),
+			"last_modified": schema.StringAttribute{
```
henryrecker-pingidentity, line 84: I think we can leave read-only modification timestamps like this out of the provider. They tend to cause issues with unexpected plans.

internal/resource/config/authenticationselector/authentication_selector_resource.go
```diff
@@ -0,0 +1,353 @@
...
+			},
+			"plugin_descriptor_ref": schema.SingleNestedAttribute{
+				Required:    true,
+				Description: "Reference to the plugin descriptor for this instance. The plugin descriptor cannot be modified once the instance is created.",
+				Attributes:  resourcelink.ToSchemaNoLocation(),
+				PlanModifiers: []planmodifier.Object{
+					objectplanmodifier.RequiresReplace(),
+				},
+			},
+			"parent_ref": schema.SingleNestedAttribute{
```
henryrecker-pingidentity, line 78: Should we allow parents? Not sure if this is the same deal as the inherited attribute we want to remove.

internal/resource/config/authenticationselector/authentication_selector_resource.go
```diff
@@ -0,0 +1,353 @@
...
+func (r *authenticationSelectorResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	var plan authenticationSelectorResourceModel
+
+	diags := req.Plan.Get(ctx, &plan)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	var hasObjectErrMap = make(map[error]bool)
```
henryrecker-pingidentity, line 198: It looks like you only ever add to this once, so there isn't really a need for the map

internal/resource/config/authenticationselector/authentication_selector_resource.go
```diff
@@ -0,0 +1,349 @@
...
+func (r *authenticationSelectorResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+
+	var plan authenticationSelectorResourceModel
+	diags := req.Plan.Get(ctx, &plan)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	var hasObjectErrMap = make(map[error]bool)
```
henryrecker-pingidentity, line 284: Same things here, the map isn't needed currently

internal/resource/config/authenticationselector/authentication_selector_resource.go
```diff
@@ -0,0 +1,349 @@
...
+		if hasErr {
+			resp.Diagnostics.AddError("Failed to create an Authentication Selector due to dependent object", err.Error())
+		}
+	}
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	createAuthenticationSelectors := client.NewAuthenticationSelector(plan.SelectorId.ValueString(), plan.Name.ValueString(), *pluginDescriptorRef, configuration)
```
henryrecker-pingidentity, line 221: We haven't done this for other resources yet but we could start using a common createAuthenticationSelectorRequest (or something) method that could reduce a little bit of duplication between Create and Update for building the required parameters.

internal/resource/config/authenticationselector/authentication_selector_resource.go
wesleymccollam: I think so.. Maybe we can remove later if needed ¯\_(ツ)_/¯

internal/resource/config/authenticationselector/authentication_selector_resource.go
wesleymccollam: nice catch

internal/resource/config/authenticationselector/authentication_selector_resource.go
wesleymccollam: I've added a JIRA case for this

internal/resource/config/authenticationselector/authentication_selector_resource.go
wesleymccollam: removed

internal/resource/config/authenticationselector/authentication_selector_resource.go
wesleymccollam: I added configuration to the map now

CHANGELOG.md
wesleymccollam: corrected

henryrecker-pingidentity (approved):

PR #201, opened 2024-02-29
fix authentication selector test for all versions

wesleymccollam:
* fix the authentication selector acceptance test to support all PingFederate versions. Previously, this was not working - see the run below.
https://github.com/pingidentity/terraform-provider-pingfederate/actions/runs/8092260678/job/22112683103

### DIFF
--- a/internal/acctest/config/authenticationselector/authentication_selector_test.go
+++ b/internal/acctest/config/authenticationselector/authentication_selector_test.go
@@ -21,7 +21,6 @@ type authenticationSelectorsResourceModel struct {
 	attributeContract                *client.AuthenticationSelectorAttributeContract
 	pluginDescriptorRef              client.ResourceLink
 	addOrUpdateAuthNContextAttribute string
-	overrideAuthNContextForFlow      string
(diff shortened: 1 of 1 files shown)


internal/acctest/config/authenticationselector/authentication_selector_test.go
```diff
@@ -127,32 +124,27 @@ resource "pingfederate_authentication_selector" "%[1]s" {
         name  = "Add or Update AuthN Context Attribute"
         value = "%[4]s"
       },
-      {
-        name  = "Override AuthN Context for Flow"
```
henryrecker-pingidentity, line 131: What's the restriction on this? We could follow the example of some other tests and add this in if it makes sense for the version being tested.

henryrecker-pingidentity (approved):

internal/acctest/config/authenticationselector/authentication_selector_test.go
wesleymccollam: added version check

henryrecker-pingidentity (approved):

PR #202, opened 2024-02-29
attribute contract nested lists to sets

### LINKED ISSUE #200: Apply Error with Token Manager
Hi Team, 

I am trying to use the resource "pingfederate_oauth_access_token_manager". The plan works perfectly and the apply also seems to create the resource however I am getting the following error after the apply: 

│ When applying changes to module.oidcmodule.pingfederate_oauth_access_token_manager.token_manager["test-app"], provider "module.oidcmodule.provider[\"registry.terraform.io/pingidentity/pingfederate\"].pingidentity" produced an
│ unexpected new value: .attribute_contract.extended_attributes[0].multi_valued: was cty.False, but now cty.True
│

The attribute contract in my resource is defined "pingfederate_oauth_access_token_manager" as: 

 attribute_contract = {
    extended_attributes = [
      {
        name         = "contract"
        multi_valued = false
      }
    ]
  }

I am also getting the following error when I add multiple extended attributes: 

╷
│
│ When applying changes to module.oidcmodule.pingfederate_oauth_access_token_manager.token_manager["test-app"], provider
│ "module.oidcmodule.provider[\"registry.terraform.io/pingidentity/pingfederate\"].pingidentity" produced an unexpected new value:
│ .attribute_contract.extended_attributes[5].name: was cty.StringVal("attribute1"), but now cty.StringVal("attribute2").
│

I think it's because it is trying to compare the resource that is created but it is seeing some inconsistencies (even if that isn't the case) . It would be much appreciated if you could fix this bug. Thank you !

wesleymccollam:
Resolves #200

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# v0.7.1 February 29th, 2024
+### BUG FIXES
+* `pingfederate_oauth_access_token_manager` resource `attribute_contract.extended_attributes` Provider produced inconsistent result after apply ([#202](https://github.com/pingidentity/terraform-provider-pingfederate/pull/202))
+
 # v0.7.0 February 28th, 2024
 ### DEPRECATED
 * `location` property in `resourceLink` object types for all Resources and DataSources. This property will be removed in a future release. ([#195](https://github.com/pingidentity/terraform-provider-pingfederate/pull/195))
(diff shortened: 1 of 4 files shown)


henryrecker-pingidentity:
We should add a test that covers this - it should fail without the change to the resource file and succeed with the change.

henryrecker-pingidentity (approved):

PR #204, opened 2024-03-01
bug fix for extended parameters

### LINKED ISSUE #203: Apply error for pingfederate_oauth_client
Hi Team,

Thank you so much for fixing the bug I mentioned yesterday. I am here to ask to fix a similar bug but, in the resource “pingfederate_oauth_client”. I am defining extended parameters like so, where I am doing a for_each on my resource: 

extended_parameters = {

param1 = {
values = [each.value.var1] 
}

param2 =  {
values = [each.value.var2] 
}

param3 = try(each.value.var3,null) == null ? {} :{
values = [each.value.var3] 
}

}

Similar to the previous issue, when the param3 is set to {} the plan and apply works but I am getting the following an error in the comparison:

│
│ When applying changes to module.oidcmodule.pingfederate_oauth_client.oidc_integration["test-app"], provider
│ "module.oidcmodule.provider[\"registry.terraform.io/pingidentity/pingfederate\"].pingidentity" produced an unexpected new value: .extended_parameters["param3"].values: │ was null, but now cty.SetValEmpty(cty.String).
│

I have certain instances where the variable var3 may not be defined, so that is why I have that if statement (the extended_parameter param3 only exists if var3 exists). Thank you so much in advance for your time and reactivity !

wesleymccollam:
Resolves #203

### DIFF
--- a/internal/resource/config/oauth/client/oauth_client_common.go
+++ b/internal/resource/config/oauth/client/oauth_client_common.go
@@ -239,9 +239,8 @@ func readOauthClientResponseCommon(ctx context.Context, r *client.Client, state,
 		extendedParametersToState, respDiags = types.MapValueFrom(ctx, extendedParametersObjAttrType, r.ExtendedParameters)
 		diags.Append(respDiags...)
 	} else {
-		if internaltypes.IsDefined(plan.ExtendedParameters) {
-			extendedParametersToState, respDiags = types.MapValueFrom(ctx, extendedParametersObjAttrType, plan.ExtendedParameters)
-			diags.Append(respDiags...)
+		if plan != nil && internaltypes.IsDefined(plan.ExtendedParameters) {
+			extendedParametersToState = plan.ExtendedParameters
 		} else {
 			extendedParametersToState = types.MapNull(extendedParametersObjAttrType)
 		}
--- a/internal/resource/config/oauth/client/oauth_client_common.go
+++ b/internal/resource/config/oauth/client/oauth_client_common.go
@@ -7,6 +7,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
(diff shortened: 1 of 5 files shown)


internal/resource/config/oauth/client/oauth_client_common.go
```diff
@@ -228,11 +234,19 @@ func readOauthClientResponseCommon(ctx context.Context, r *client.Client, state
...
-	extendedParametersAttrType["values"] = types.SetType{ElemType: types.StringType}
-	extendedParametersObjAttrType := types.ObjectType{AttrTypes: extendedParametersAttrType}
-	extendedParametersToState, respDiags := types.MapValueFrom(ctx, extendedParametersObjAttrType, r.ExtendedParameters)
-	diags.Append(respDiags...)
+	var extendedParametersToState basetypes.MapValue
+	if r.ExtendedParameters != nil {
+		extendedParametersToState, respDiags = types.MapValueFrom(ctx, extendedParametersObjAttrType, r.ExtendedParameters)
+		diags.Append(respDiags...)
+	} else {
+		if internaltypes.IsDefined(plan.ExtendedParameters) {
```
henryrecker-pingidentity, line 242: I'd add a nil check for plan here, and just pass in nil for the plan in the data source.

internal/resource/config/oauth/client/oauth_client_common.go
```diff
@@ -228,11 +234,19 @@ func readOauthClientResponseCommon(ctx context.Context, r *client.Client, state
...
+			extendedParametersToState, respDiags = types.MapValueFrom(ctx, extendedParametersObjAttrType, plan.ExtendedParameters)
```
henryrecker-pingidentity, line 243: I don't think you need to use MapValueFrom here - plan.ExtendedParameters should already be the Map you need.

internal/resource/config/oauth/client/oauth_client_common.go
wesleymccollam: nil check for plan is now added

internal/resource/config/oauth/client/oauth_client_common.go
wesleymccollam: updated to below
`if plan != nil && internaltypes.IsDefined(plan.ExtendedParameters) {
			extendedParametersToState = plan.ExtendedParameters`

PR #215, opened 2024-03-13
pingfederate_oauth_access_token_mapping attribute_contract_fulfillment bug fix

### LINKED ISSUE #210: Apply Error with resource pingfederate_oauth_access_token_mapping
Hi Team, 

I am getting a similar error to my previous requests, but on the resource pingfederate_oauth_access_token_mapping. When I define the  attribute_contract_fulfillment as follows: 

  attribute_contract_fulfillment = {
       "amr" = {
           source = {
               type = "NO_MAPPING"
            }
        }, 
     "username" = {
          source = {
               type = "NO_MAPPING"
            }
        },
}

The resource does get created with the attribute_contract_fulfillment define but I am getting the following error after an apply: 

│
│ When applying changes to pingfederate_oauth_access_token_mapping.token_mappings, provider
│ "provider[\"registry.terraform.io/pingidentity/pingfederate\"].pingidentity" produced an unexpected new value:
│ .attribute_contract_fulfillment["amr"].value: was null, but now cty.StringVal("").
│
│
│ When applying changes to pingfederate_oauth_access_token_mapping.token_mappings, provider
│ "provider[\"registry.terraform.io/pingidentity/pingfederate\"].pingidentity" produced an unexpected new value:
│ .attribute_contract_fulfillment["username"].value: was null, but now cty.StringVal("").
│

Once again Thank you  for your reactivity !

wesleymccollam:
Resolves #210

### DIFF
--- a/internal/resource/config/oauth/accesstokenmapping/oauth_access_token_mappings_resource.go
+++ b/internal/resource/config/oauth/accesstokenmapping/oauth_access_token_mappings_resource.go
@@ -93,7 +93,7 @@ func (r *oauthAccessTokenMappingsResource) Schema(ctx context.Context, req resou
 				Attributes: resourcelink.ToSchemaNoLocation(),
 			},
 			"attribute_sources":              attributesources.ToSchema(0, true, false),
-			"attribute_contract_fulfillment": attributecontractfulfillment.ToSchema(true, false, false),
+			"attribute_contract_fulfillment": attributecontractfulfillment.ToSchema(true, false, true),
 			"issuance_criteria":              issuancecriteria.ToSchema(),
 		},
 	}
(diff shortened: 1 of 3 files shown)


internal/resource/config/oauth/accesstokenmapping/oauth_access_token_mappings_resource.go
```diff
@@ -93,7 +93,7 @@ func (r *oauthAccessTokenMappingsResource) Schema(ctx context.Context, req resou
 				Attributes: resourcelink.ToSchemaNoLocation(),
 			},
 			"attribute_sources":              attributesources.ToSchema(0, true, false),
-			"attribute_contract_fulfillment": attributecontractfulfillment.ToSchema(true, false, false),
+			"attribute_contract_fulfillment": attributecontractfulfillment.ToSchema(true, false, true),
```
henryrecker-pingidentity, line 96: It seems like from the bug description we should be setting a default empty string value for the "value" attribute. Does this method set a default?

wesleymccollam: Added string default

henryrecker-pingidentity (approved):

PR #216, opened 2024-03-13
correct required fields in pingfederate_oauth_auth_server_settings resource

wesleymccollam:


internal/acctest/config/oauth/authserversettings/oauth_auth_server_settings_test.go
```diff
@@ -83,6 +79,25 @@ func TestAccOauthAuthServerSettings(t *testing.T) {
 }
 
 func testAccOauthAuthServerSettings(resourceName string, resourceModel oauthAuthServerSettingsResourceModel, includeAllAttributes bool) string {
+	addUpdatedResourceModelFields := []string{}
+	if resourceModel.bypassActivationCodeConfirmation == true {
+		addUpdatedResourceModelFields = append(addUpdatedResourceModelFields, "bypass_activation_code_confirmation = true")
```
henryrecker-pingidentity, line 84: Seems like these values should come from the resourceModel rather than being hard-coded

internal/resource/config/oauth/authserversettings/oauth_auth_server_settings_resource.go
```diff
@@ -380,18 +382,24 @@ func (r *oauthAuthServerSettingsResource) Schema(ctx context.Context, req resour
...
 				Validators: []validator.String{
 					configvalidators.StartsWith("/"),
 				},
 			},
 			"pending_authorization_timeout": schema.Int64Attribute{
 				Description: "The 'device_code' and 'user_code' timeout, in seconds.",
-				Required:    true,
+				Optional:    true,
+				Computed:    true,
+				Default:     int64default.StaticInt64(600),
```
henryrecker-pingidentity, line 396: Where do these defaults come from?

internal/resource/config/oauth/authserversettings/oauth_auth_server_settings_resource.go
```diff
@@ -89,7 +89,9 @@ func (r *oauthAuthServerSettingsResource) Schema(ctx context.Context, req resour
 		Attributes: map[string]schema.Attribute{
 			"default_scope_description": schema.StringAttribute{
 				Description: "The default scope description.",
-				Required:    true,
+				Optional:    true,
+				Computed:    true,
+				Default:     stringdefault.StaticString(""),
```
henryrecker-pingidentity, line 94: Do we need a default here, or would it be fine to just leave this optional and not computed?

internal/resource/config/oauth/authserversettings/oauth_auth_server_settings_resource.go
wesleymccollam: PingFederate GET request after spinning a new instance

internal/resource/config/oauth/authserversettings/oauth_auth_server_settings_resource.go
wesleymccollam: The default value returned from PingFederate is an empty string.

internal/resource/config/oauth/authserversettings/oauth_auth_server_settings_resource.go
wesleymccollam: Also in the response when the value is not sent in.

internal/acctest/config/oauth/authserversettings/oauth_auth_server_settings_test.go
wesleymccollam: this is now updated

internal/resource/config/oauth/authserversettings/oauth_auth_server_settings_resource.go
henryrecker-pingidentity: Ok thanks. I forgot this was a settings resource as well.

henryrecker-pingidentity (approved):

PR #217, opened 2024-03-14
url_validator ignores empty string

wesleymccollam:
- url_validator config validator ignores empty string
- test coverage where bug discovered


henryrecker-pingidentity (approved):
Makes sense, looks good to me

PR #221, opened 2024-03-25
pingfederate_oauth_client multiple bug fixes - see changelog

wesleymccollam:


CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# v0.9.0 (Unreleased)
+#### BUG FIXES
+* `pingfederate_oauth_client` resource has the following bugs resolved ([#221](https://github.com/pingidentity/terraform-provider-pingfederate/pull/221)):
+  - `client_id` now correctly forces a replacement of the resource. Previously if the resource was created, then the `client_id` value was changed, would result in `"Error deleting client. Client <client_id> not found"}`
+  - `Error: client_auth.secret cannot be empty when "CLIENT_CREDENTIALS" is included in grant_types.` when supplying sensitive attribute
```
henryrecker-pingidentity, line 5: It would be better to describe the bug fixes here than copying the error text. Like "fixed unexpected new value error when setting restrict_scopes to false" or whatever.

internal/resource/config/oauth/client/oauth_client_common.go
```diff
@@ -45,8 +45,9 @@ var (
 		"ping_access_logout_capable":                  types.BoolValue(false),
 		"logout_uris":                                 types.SetNull(types.StringType),
 		"pairwise_identifier_user_type":               types.BoolValue(false),
-		"logout_mode":                                 types.StringUnknown(),
-		"back_channel_logout_uri":                     types.StringUnknown(),
+		"sector_identifier_uri":                       types.StringNull(),
```
henryrecker-pingidentity, line 48: Was this just missing before on the default?

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -889,13 +892,9 @@ func (r *oauthClientResource) ValidateConfig(ctx context.Context, req resource.V
 		if grantTypeVal == "CLIENT_CREDENTIALS" {
 			if clientAuthDefined {
 				clientAuthType := clientAuthAttributes["type"].(types.String).ValueString()
-				clientAuthSecret := clientAuthAttributes["secret"].(types.String).ValueString()
 				if clientAuthType != "NONE" && clientAuthType != "SECRET" {
 					resp.Diagnostics.AddError("client_auth.type must be set to \"SECRET\" when \"CLIENT_CREDENTIALS\" is included in grant_types.", "")
 				}
-				if clientAuthSecret == "" {
```
henryrecker-pingidentity, line 896: What was the issue here?

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -1041,6 +1044,11 @@ func (r *oauthClientResource) ModifyPlan(ctx context.Context, req resource.Modif
 		}
 	}
 
+	if plan.AllowAuthenticationApiInit.ValueBool() && internaltypes.IsDefined(plan.RestrictScopes) && !plan.RestrictScopes.ValueBool() {
```
henryrecker-pingidentity, line 1047: If plan.RestrictScopes is non-Unknown you aren't allowed to change the value here, since the true value won't match the config.

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -1182,13 +1191,10 @@ func addOptionalOauthClientFields(ctx context.Context, addRequest *client.Client
...
 	}
 
-	if internaltypes.IsDefined(plan.RestrictScopes) {
-		addRequest.RestrictScopes = plan.RestrictScopes.ValueBoolPointer()
-		if *plan.RestrictScopes.ValueBoolPointer() && internaltypes.IsDefined(plan.RestrictedScopes) {
-			var slice []string
-			plan.RestrictedScopes.ElementsAs(ctx, &slice, false)
-			addRequest.RestrictedScopes = slice
-		}
+	if *plan.RestrictScopes.ValueBoolPointer() && internaltypes.IsDefined(plan.RestrictedScopes) {
```
henryrecker-pingidentity, line 1194: It looks like you could just use ValueBool here

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -1182,13 +1191,10 @@ func addOptionalOauthClientFields(ctx context.Context, addRequest *client.Client
...
+		var slice []string
+		plan.RestrictedScopes.ElementsAs(ctx, &slice, false)
```
henryrecker-pingidentity, line 1196: Would it make more sense for this if logic to just be part of validate config instead? It seems like silently ignoring the restricted_scopes value rather than printing an explicit error is harder for a user to understand.

internal/resource/config/oauth/client/oauth_client_common.go
wesleymccollam: yes, was missing.

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: I'm reworking this now, this is incorrect

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -1041,6 +1050,11 @@ func (r *oauthClientResource) ModifyPlan(ctx context.Context, req resource.Modif
 		}
 	}
 
+	if plan.AllowAuthenticationApiInit.ValueBool() && !internaltypes.IsDefined(plan.RestrictScopes) {
+		plan.RestrictScopes = types.BoolValue(true)
```
henryrecker-pingidentity, line 1054: This can only be done if plan.RestrictScopes is specifically Unknown

internal/resource/config/oauth/client/oauth_client_resource.go
```diff
@@ -1155,6 +1169,10 @@ func addOptionalOauthClientFields(ctx context.Context, addRequest *client.Client
 	addRequest.JwtSecuredAuthorizationResponseModeEncryptionAlgorithm = plan.JwtSecuredAuthorizationResponseModeEncryptionAlgorithm.ValueStringPointer()
 	addRequest.JwtSecuredAuthorizationResponseModeContentEncryptionAlgorithm = plan.JwtSecuredAuthorizationResponseModeContentEncryptionAlgorithm.ValueStringPointer()
 	addRequest.RequireDpop = plan.RequireDpop.ValueBoolPointer()
+	addRequest.RestrictScopes = plan.RestrictScopes.ValueBoolPointer()
+	var slice []string
```
henryrecker-pingidentity, line 1173: I'd give this a more specific name

internal/acctest/config/oauth/client/oauth_client_test.go
```diff
@@ -90,7 +90,7 @@ func TestAccOauthClient(t *testing.T) {
 		description:                        pointers.String("updatedDescription"),
 		logoUrl:                            pointers.String("https://example.com"),
 		redirectUris:                       []string{"https://example.com"},
-		allowAuthenticationApiInit:         pointers.Bool(false),
+		allowAuthenticationApiInit:         pointers.Bool(true),
```
henryrecker-pingidentity, line 93: At some point we may want to add tests to cover expected errors such as errors from modifyplan

wesleymccollam: I added a case for this internally.

internal/resource/config/oauth/client/oauth_client_resource.go
wesleymccollam: updated

henryrecker-pingidentity (approved):

PR #229, opened 2024-04-03
acceptance tests - state coverage

wesleymccollam:


internal/resource/config/serversettings/server_settings_resource.go
```diff
@@ -431,7 +431,10 @@ func (r *serverSettingsResource) Schema(ctx context.Context, req resource.Schema
 						Description: "If supplied, the Source ID value entered here is used for SAML 1.x, instead of being derived from the SAML 1.x Issuer/Audience.",
 						Computed:    true,
 						Optional:    true,
-						Default:     stringdefault.StaticString(""),
+						Validators: []validator.String{
+							stringvalidator.LengthBetween(40, 40),
```
henryrecker-pingidentity, line 435: Where does this come from?

internal/resource/config/oauth/authserversettings/oauth_auth_server_settings_resource.go
```diff
@@ -285,6 +286,21 @@ func (r *oauthAuthServerSettingsResource) Schema(ctx context.Context, req resour
 				Description: "The grant types that the OAuth AS can reuse rather than creating a new grant for each request. Only 'IMPLICIT' or 'AUTHORIZATION_CODE' or 'RESOURCE_OWNER_CREDENTIALS' are valid grant types.",
 				Computed:    true,
 				Optional:    true,
+				Validators: []validator.Set{
+					setvalidator.ValueStringsAre(
+						stringvalidator.OneOf(
```
henryrecker-pingidentity, line 291: This doesn't align with the description of the attribute

internal/resource/config/oauth/authserversettings/oauth_auth_server_settings_resource.go
```diff
@@ -269,7 +270,7 @@ func (r *oauthAuthServerSettingsResource) Schema(ctx context.Context, req resour
 				Description: "The roll refresh token values default policy. The default value is true.",
 				Computed:    true,
 				Optional:    true,
-				Default:     booldefault.StaticBool(true),
+				Default:     booldefault.StaticBool(false),
```
henryrecker-pingidentity, line 273: This doesn't align with the description of the attribute. Why was this changed?

internal/acctest/acctest.go
```diff
@@ -266,6 +277,51 @@ func AddIdHcl(idKey, id string) string {
 	return ""
 }
 
+func ReturnStateMismatchError(stateKey, stateValue, configValue string) error {
+	return fmt.Errorf("state value for %s does not match config value. State: %s, Found: %s", stateKey, stateValue, configValue)
+}
+
+func VerifyStateAttributeValue(stateAttributes map[string]string, stateKeyValue string, configValue interface{}) error {
```
henryrecker-pingidentity, line 284: Do we need this custom method? I thought the testing framework had its own built-in methods to verify state that we use in a few tests.

internal/acctest/acctest.go
```diff
@@ -266,6 +277,51 @@ func AddIdHcl(idKey, id string) string {
...
+		}
+	}
+
+	return nil
+}
+
+func VerifyStateAttributeSlice(stateAttributes map[string]string, stateKeyValue string, configValue []string) error {
+	configValueToTest := StringSliceToTerraformString(configValue)
+
+	if stateAttributes[stateKeyValue] != configValueToTest && (stateAttributes[stateKeyValue] != "" && configValueToTest == "[]") {
```
henryrecker-pingidentity, line 316: This logic doesn't seem right - any time `configValueToTest` is not exactly "[]" this returns an error.

internal/acctest/acctest.go
```diff
@@ -266,6 +277,51 @@ func AddIdHcl(idKey, id string) string {
...
+		err := ReturnStateMismatchError(stateKeyValue, stateAttributes[stateKeyValue], configValueToTest)
+		if err != nil {
```
henryrecker-pingidentity, line 318: There's no need to check for non-nil here

internal/resource/config/serversettings/server_settings_resource.go
wesleymccollam: Received an error stating this when testing the attributes - the length must be exactly 40 characters.

internal/resource/config/oauth/authserversettings/oauth_auth_server_settings_resource.go
wesleymccollam: Due to a result of my testing after a fresh spin. However, it could be based on the current config. I'll change it back and trust the description.

henryrecker-pingidentity (approved):

PR #230, opened 2024-04-04
add missing data.json.subst's for nightly tests

wesleymccollam:


henryrecker-pingidentity (commented):
Any way we can have logic to reuse these rather than duplicating? Maybe one profile per major-minor version if that's all we really need?

henryrecker-pingidentity (approved):

PR #232, opened 2024-04-16
Authentication policy resource and data source

wesleymccollam:


internal/acctest/config/authenticationpoliciespolicy/authentication_policies_policy_test.go
```diff
@@ -0,0 +1,290 @@
...
+				),
+			},
+			{
+				// Test importing the resource
+				Config:            testAccAuthenticationPoliciesPolicyComplex(resourceName),
+				ResourceName:      "pingfederate_authentication_policies_policy." + resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+			{
```
henryrecker-pingidentity, line 59: This seems to be missing the deletion check

internal/acctest/config/authenticationpoliciespolicy/authentication_policies_policy_test.go
```diff
@@ -0,0 +1,290 @@
...
+				Config: testAccAuthenticationPoliciesPolicySimple(resourceName),
+				Check:  testAccCheckExpectedAuthenticationPoliciesPolicyAttributes(resourceName, false),
+			},
+			// Test a more complex policy
+			{
+				Config: testAccAuthenticationPoliciesPolicyComplex(resourceName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckExpectedAuthenticationPoliciesPolicyAttributes(resourceName, true),
+					resource.TestCheckResourceAttr(fmt.Sprintf("pingfederate_authentication_policies_policy.%s", resourceName), "enabled", "true"),
```
henryrecker-pingidentity, line 34: I don't think we need checks like this for anything we explicitly set in the HCL, we need them for things that are computed that we want to verify are in place.

internal/resource/common/attributemapping/attribute_mapping.go
```diff
@@ -34,7 +38,36 @@ func ToSchema() schema.SingleNestedAttribute {
 			"attribute_sources":              attributesources.ToSchema(0, false, true),
 			"issuance_criteria":              issuancecriteria.ToSchema(),
 		},
-		Required:    true,
+		Required:    false,
```
henryrecker-pingidentity, line 41: Should this be a parameter? This change will affect any other resources using attribute mappings.

internal/resource/common/policyaction/policy_action_client_struct.go
```diff
@@ -68,6 +77,15 @@ func ClientStruct(object types.Object) (*client.PolicyActionAggregation, error)
 		if err != nil {
 			return nil, err
 		}
+
+		attributeSources := fragmentPolicyAction.(types.Object).Attributes()["fragment_mapping"].(types.Object).Attributes()["attribute_sources"]
+		if attributeSources != nil {
```
henryrecker-pingidentity, line 82: It would be more go-like if these checks used a second `ok` parameter to verify that "attribute_sources" is in the map, rather than the nil check here

internal/resource/config/authenticationpolicies/policy/authentication_policies_policy_resource.go
```diff
@@ -0,0 +1,252 @@
...
+		return
+	}
+
+	providerCfg := req.ProviderData.(internaltypes.ResourceConfiguration)
+	r.providerConfig = providerCfg.ProviderConfig
+	r.apiClient = providerCfg.ApiClient
+
+}
+
+func readAuthenticationPolicyResponse(ctx context.Context, r *client.AuthenticationPolicyTree, state *authenticationPoliciesPolicyModel) diag.Diagnostics {
```
henryrecker-pingidentity, line 99: In other resources/datasources we have put this method and the model in a separate _common.go file, since it is being used both here and in the data source

internal/resource/config/authenticationpolicies/policy/authentication_policies_policy_resource.go
```diff
@@ -0,0 +1,252 @@
...
+			},
+			"authentication_api_application_ref": resourcelink.CompleteSingleNestedAttribute(true, false, false,
+				"Authentication API Application Id to be used in this policy branch. If the value is not specified, no Authentication API Application will be used."),
+			"enabled": schema.BoolAttribute{
+				Optional:    true,
+				Computed:    true,
+				Default:     booldefault.StaticBool(true),
+				Description: "Whether or not this authentication policy tree is enabled. Default is true.",
+			},
+			"root_node": authenticationpolicytreenode.ToSchema(),
```
henryrecker-pingidentity, line 69: I think the description may differ for this node vs the root node of a fragment, so we may need a parameter for that.

internal/resource/config/authenticationpolicies/policy/authentication_policies_policy_resource.go
```diff
@@ -0,0 +1,252 @@
...
+		return
+	}
+	httpResp, err := r.apiClient.AuthenticationPoliciesAPI.DeletePolicy(config.ProviderBasicAuthContext(ctx, r.providerConfig), state.PolicyId.ValueString()).Execute()
+	if err != nil && (httpResp == nil || httpResp.StatusCode != 404) {
+		config.ReportHttpError(ctx, &resp.Diagnostics, "An error occurred while deleting an Authentication Policy", err, httpResp)
+	}
+}
+
+func (r *authenticationPoliciesPolicyResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
```
henryrecker-pingidentity, line 251: This should be policy_id

internal/resource/config/authenticationpolicies/policy/authentication_policies_policy_resource.go
```diff
@@ -0,0 +1,252 @@
...
+func (r *authenticationPoliciesPolicyResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+	var state authenticationPoliciesPolicyModel
+
+	diags := req.State.Get(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	policyResponse, httpResp, err := r.apiClient.AuthenticationPoliciesAPI.GetPolicy(config.ProviderBasicAuthContext(ctx, r.providerConfig), state.Id.ValueString()).Execute()
```
henryrecker-pingidentity, line 182: This method should use PolicyId, the normal Id is only computed for terraform testing purposes.

internal/resource/common/authenticationpolicytreenode/authentication_policy_tree_node_state.go
```diff
@@ -44,8 +44,8 @@ func childrenAttrTypes(depth int) types.ListType {
 	return childrenAttrTypesByDepth[depth]
 }
 
-func State(ctx context.Context, node *client.AuthenticationPolicyTreeNode) (types.Object, diag.Diagnostics) {
-	return recursiveState(ctx, node, 1, getRootNodeAttrTypes())
+func ToState(ctx context.Context, node *client.AuthenticationPolicyTreeNode, depth int) (types.Object, diag.Diagnostics) {
```
henryrecker-pingidentity, line 47: Why was depth added as a parameter here? This method always expects to be called from the root node, which has a depth of 1.

internal/resource/common/policyaction/policy_action_schema.go
```diff
@@ -81,7 +81,7 @@ func commonAttributeRulesAttr() schema.Attribute {
 
 func apcMappingPolicyActionSchema() schema.SingleNestedAttribute {
 	attrs := commonPolicyActionSchema()
-	attrs["attribute_mapping"] = attributemapping.ToSchema()
+	attrs["attribute_mapping"] = attributemapping.ToSchema(false)
```
henryrecker-pingidentity, line 84: This one is required per the API doc page

internal/resource/common/policyaction/policy_action_schema.go
```diff
@@ -188,13 +188,13 @@ func fragmentPolicyActionSchema() schema.SingleNestedAttribute {
...
 	attrs := commonPolicyActionSchema()
-	attrs["inbound_mapping"] = attributemapping.ToSchema()
+	attrs["inbound_mapping"] = attributemapping.ToSchema(false)
 	attrs["local_identity_ref"] = schema.SingleNestedAttribute{
 		Attributes:  resourcelink.ToSchema(),
 		Required:    true,
 		Description: "A reference to a resource.",
 	}
-	attrs["outbound_attribute_mapping"] = attributemapping.ToSchema()
+	attrs["outbound_attribute_mapping"] = attributemapping.ToSchema(false)
```
henryrecker-pingidentity, line 197: This one is also required

internal/resource/common/policyaction/policy_action_schema.go
wesleymccollam: whoops, nice catch

internal/acctest/config/authenticationpoliciespolicy/authentication_policies_policy_test.go
wesleymccollam: added

internal/resource/common/policyaction/policy_action_schema.go
wesleymccollam: these are updated

PR #263, opened 2024-06-17
pingfederate_oauth_idp_adapter_mapping resource

henryrecker-pingidentity:


examples/resources/pingfederate_oauth_idp_adapter_mapping/resource.tf
```diff
@@ -0,0 +1,17 @@
+resource "pingfederate_oauth_idp_adapter_mapping" "oauthIdpAdapterMapping" {
```
henryrecker-pingidentity, line 1: I'm unsure if this example is very good - I didn't find any examples of this resource in facile or in a sample PF data.json

internal/resource/config/oauth/idpadaptermappings/oauth_idp_adapter_mapping_resource.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+func (r *oauthIdpAdapterMappingResource) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
+	var plan *oauthIdpAdapterMappingResourceModel
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	if plan == nil {
+		return
+	}
+	// The idp adapter object will be configured based on the provided mapping_id
```
henryrecker-pingidentity, line 17: This, strangely, is the opposite of how the authentication policy contract mapping version works. In that resource, the mapping_id is computed based on what is set in the idp_adapter_ref. This resource works exactly the opposite.

Maybe it makes sense for the provider to force one way or the other here?

wesleymccollam (approved):

patrickcping (commented):
PDI-1898

PR #264, opened 2024-06-20
pingfederate_idp_to_sp_adapter_mapping resource

henryrecker-pingidentity:


examples/resources/pingfederate_idp_to_sp_adapter_mapping/resource.tf
henryrecker-pingidentity, line 1: I couldn't find any real examples of this one, so we probably need a better example here.

patrickcping (approved):
LGTM 🚀 🚀

PR #265, opened 2024-06-21
pingfederate_sp_adapter resource

henryrecker-pingidentity:


internal/resource/config/sp/adapters/sp_adapter_resource_gen.go
```diff
@@ -0,0 +1,428 @@
...
+			},
+			"configuration": pluginconfiguration.ToSchema(),
+			"name": schema.StringAttribute{
+				Required:    true,
+				Description: "The plugin instance name. The name can be modified once the instance is created.<br>Note: Ignored when specifying a connection's adapter override.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"parent_ref": schema.SingleNestedAttribute{
```
henryrecker-pingidentity, line 119: Should we remove these parent_ref attrs from resources in the provider? I'm not sure they make sense in a terraform context.

wesleymccollam: Good question!

wesleymccollam (approved):

internal/resource/config/sp/adapters/sp_adapter_resource_gen.go
patrickcping: We'll leave them in - I can see a benefit here, because you can opt to leave all the fields out of the configuration and they will still inherit from the parent

patrickcping (commented):
PDI-1999

PR #281, opened 2024-07-08
pingfederate_idp_sts_request_parameters_contract resource

henryrecker-pingidentity:


examples/resources/pingfederate_idp_sts_request_parameters_contract/resource.tf
```diff
@@ -0,0 +1,9 @@
+resource "pingfederate_idp_sts_request_parameters_contract" "requestParametersContract" {
```
henryrecker-pingidentity, line 1: I wasn't able to find a realistic example for this resource.

wesleymccollam (approved):

examples/resources/pingfederate_idp_sts_request_parameters_contract/resource.tf
patrickcping: this is fine

patrickcping (approved):
LGTM 🚀 🚀

PR #282, opened 2024-07-08
pingfederate_metadata_url resource

henryrecker-pingidentity:


examples/resources/pingfederate_metadata_url/resource.tf
```diff
@@ -0,0 +1,10 @@
+resource "pingfederate_metadata_url" "metadataUrl" {
```
henryrecker-pingidentity, line 1: I didn't find a good example for this one

internal/resource/config/metadataurls/metadata_url_resource.go
```diff
@@ -0,0 +1,45 @@
...
+		// return the id value in the x509 attribute.
+		// Note that this method assumes the response CertView has already been set in state.
+		idAttr := types.StringNull()
+		if internaltypes.IsDefined(state.CertView) {
+			idAttr = state.CertView.Attributes()["id"].(types.String)
+		}
+		x509FileValue, diags = types.ObjectValue(x509FileAttrTypes, map[string]attr.Value{
+			"crypto_provider":     types.StringPointerValue(response.X509File.CryptoProvider),
+			"file_data":           fileDataAttr,
+			"formatted_file_data": types.StringValue(response.X509File.FileData),
```
henryrecker-pingidentity, line 38: PF formats the file_data you supply in a (as far as I can tell) unpredictable way - it will shift around the newlines in the PEM contents and it will also add the header/footer as necessary. I ended up adding this as a separate attribute to handle the response from PF since it is likely to not match what the user supplies in HCL.

wesleymccollam (approved):

patrickcping (commented):
PDI-1939

PR #283, opened 2024-07-09
Add captcha_provider_settings resource

PingDavidR:
- closes PDI-1681


internal/resource/config/captchaproviders/settings/captcha_provider_settings_resource.go
```diff
@@ -0,0 +1,13 @@
+package captchaproviderssettings
+
+import (
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+func (m *captchaProviderSettingsResourceModel) setNullObjectValues() {
```
henryrecker-pingidentity, line 8: This shouldn't be necessary after recent updates to the code generator - it should be using the emptyModel method that gets generated in the resource_gen file.

internal/acctest/config/captchaproviders/settings/captcha_provider_settings_resource_gen_test.go
```diff
@@ -0,0 +1,67 @@
...
+		Steps: []resource.TestStep{
+			{
+				// Set with the initial model
+				Config: captchaProviderSettings_BeforeHCL(),
+			},
+			{
+				// Flip to the updated model
+				Config: captchaProviderSettings_AfterHCL(),
+			},
+			{
```
henryrecker-pingidentity, line 31: If you want it's probably fine to even delete this third step, since flipping again isn't really testing anything terraform-specific. Up to you.

internal/resource/config/captchaproviders/settings/captcha_provider_settings_resource.go
PingDavidR: fixed

internal/acctest/config/captchaproviders/settings/captcha_provider_settings_resource_gen_test.go
PingDavidR: putting it back the way it was in step 3 just in case we end up with a dependency test somewhere needing the original default in place for some odd reason

henryrecker-pingidentity (approved):

patrickcping (commented):
PDI-1938

PR #291, opened 2024-07-15
pingfederate_secret_manager resource

henryrecker-pingidentity:


examples/resources/pingfederate_secret_manager/resource.tf
henryrecker-pingidentity, line 1: I didn't find an example of this resource, so I just put the absolute minimum configuration for a CyberArkCredentialProvider here.

patrickcping (approved):
LGTM 🚀 🚀

PR #295, opened 2024-07-17
pingfederate_service_authentication resource

henryrecker-pingidentity:


wesleymccollam (approved):

internal/resource/config/serviceauthentication/service_authentication_resource_gen.go
```diff
@@ -0,0 +1,359 @@
...
+						Description: "Id of the service.",
+					},
+					"shared_secret": schema.StringAttribute{
+						Optional:    true,
+						Sensitive:   true,
+						Description: "Shared secret for the service.",
+					},
+				},
+				Optional:    true,
+				Description: "(Deprecated) Connection management service. Remove the JSON field to deactivate the connection management service.",
```
henryrecker-pingidentity, line 86: A couple of these attributes are described as deprecated in PF's docs. Should we just not include them in the provider from the start?

patrickcping: Ideally we should leave them out of the provider.  PF is following a deprecation path to encourage customers and code clients to migrate their functionality and remove those functions, so they can be safely removed from PF later.

However, because we're offering different contexts of the provider based on different product versions, we have additional complexity.  If the feature is deprecated on all of our supported versions, we should remove.  If not, then we may need extra validation (and extra documentation) to warn on the versions where the feature is deprecated, but not warn where the version being used has it not deprecated.

This means we might end up deprecated and removing fields as we un-support older versions of PF, which should also happen on major version releases of the provider

patrickcping (commented):
PDI-2005

PR #313, opened 2024-08-01
pingfederate_keypairs_signing_key resource

henryrecker-pingidentity:


examples/resources/pingfederate_keypairs_signing_key/import.sh
henryrecker-pingidentity, line 1: There might be an argument for not even supporting import on this resource - it's really not at all usable to import this as a resource. You'd get all the value by just using a data source.

henryrecker-pingidentity: I ended up just going ahead and removing import for now, but I can add it back if we feel it is useful

wesleymccollam (approved):

wesleymccollam (approved):

patrickcping (commented):
PDI-1950

PR #330, opened 2024-08-13
oauth_client_registration_policy resource

henryrecker-pingidentity:


examples/resources/pingfederate_oauth_client_registration_policy/resource.tf
```diff
@@ -0,0 +1,39 @@
+resource "pingfederate_oauth_client_registration_policy" "registrationPolicy" {
```
henryrecker-pingidentity, line 1: I didn't find an  example for this one - this is the default configuration for this type when created in the UI

wesleymccollam (approved):

patrickcping (approved):
LGTM 🚀 🚀

examples/resources/pingfederate_oauth_client_registration_policy/resource.tf
patrickcping: This works

PR #342, opened 2024-08-20
sp_idp_connection resource

wesleymccollam:


internal/resource/config/sp/idpconnection/sp_idp_connection_resource.go
```diff
@@ -0,0 +1,4529 @@
...
+									MarkdownDescription: "A list of mappings from attribute names to their fulfillment values.",
+								},
+								"attribute_sources": attributesources.ToSchema(0, false),
+								"authentication_policy_contract_ref": schema.SingleNestedAttribute{
+									Attributes:          resourcelink.ToSchema(),
+									Required:            true,
+									Description:         "A reference to a resource.",
+									MarkdownDescription: "A reference to a resource.",
+								},
+								"issuance_criteria": schema.SingleNestedAttribute{
```
henryrecker-pingidentity, line 1419: This could also use the common package

internal/resource/config/sp/idpconnection/sp_idp_connection_resource.go
```diff
@@ -0,0 +1,4529 @@
...
+						Optional:            true,
+						Description:         "The settings used to specify how and when to provision user accounts.",
+						MarkdownDescription: "The settings used to specify how and when to provision user accounts.",
+					},
+					"message_customizations": schema.SetNestedAttribute{
+						NestedObject: schema.NestedAttributeObject{
+							Attributes: map[string]schema.Attribute{
+								"context_name": schema.StringAttribute{
+									Optional:            true,
+									Description:         "The context in which the customization will be applied. Depending on the connection type and protocol, this can either be 'assertion', 'authn-response' or 'authn-request'.",
```
henryrecker-pingidentity, line 1935: Should this have an enum validator?

internal/resource/config/sp/idpconnection/sp_idp_connection_resource.go
```diff
@@ -0,0 +1,4529 @@
...
+						NestedObject: schema.NestedAttributeObject{
+							Attributes: map[string]schema.Attribute{
+								"attribute_contract_fulfillment": schema.MapNestedAttribute{
+									NestedObject: schema.NestedAttributeObject{
+										Attributes: map[string]schema.Attribute{
+											"source": schema.SingleNestedAttribute{
+												Attributes: map[string]schema.Attribute{
+													"id": schema.StringAttribute{
+														Optional:            true,
+														Description:         "The attribute source ID that refers to the attribute source that this key references. In some resources, the ID is optional and will be ignored. In these cases the ID should be omitted. If the source type i
```
henryrecker-pingidentity, line 3346: Use backtick formatting for the regex. There's also a "configvalidators.PingFederateId" validator that should be used with that regex for validation.

henryrecker-pingidentity, line 3390: This line isn't needed since this ModifyPlan doesn't actually modify the plan at all, it just throws errors

henryrecker-pingidentity, line 3647: Is this supposed to be set to null?

henryrecker-pingidentity, line 3659: This if isn't needed, the for loop will handle it (even if it is nil)

henryrecker-pingidentity, line 3660: It looks like this won't handle if the server returns a new cert in addition to the ones in the plan

henryrecker-pingidentity, line 3722: should this just be `password := ""`? Either that or just `var password string`.

henryrecker-pingidentity, line 3758: Same comments apply here

henryrecker-pingidentity, line 3907: This one might require more creativity. You might have to make a copy of the configuration or something like that.

henryrecker-pingidentity, line 4058: You'll need to give these a default value, or there will be errors about the plugin returning an invalid type if AttributeSources ever is nil. Same with issuance criteria below

internal/resource/config/sp/idpconnection/sp_idp_connection_resource.go
```diff
@@ -0,0 +1,4529 @@
...
+							Description: "A Set of values",
+						},
+					},
+				},
+				Optional:    true,
+				Description: "Extended Properties allows to store additional information for IdP/SP Connections. The names of these extended properties should be defined in /extendedProperties.",
+			},
+			"idp_browser_sso": schema.SingleNestedAttribute{
+				Attributes: map[string]schema.Attribute{
+					"adapter_mappings": schema.ListNestedAttribute{
```
henryrecker-pingidentity, line 1073: Could this be a set?

internal/acctest/config/sp/idpconnection/sp_idp_connection_resource_gen_test.go
```diff
@@ -0,0 +1,548 @@
...
+			},
+			{
+				// Test importing the resource
+				Config:            spIdpConnection_CompleteHCL(),
+				ResourceName:      "pingfederate_sp_idp_connection.example",
+				ImportStateId:     spIdpConnectionConnectionId,
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{
+					"credentials.certs.0.cert_view",
```
henryrecker-pingidentity, line 84: I would expect that cert_view could import as normal

internal/acctest/config/sp/idpconnection/sp_idp_connection_resource_gen_test.go
```diff
@@ -0,0 +1,548 @@
...
+      x509_file = {
+        id = "4qrossmq1vxa4p836kyqzp48h"
+        file_data = "MIIDOjCCAiICCQCjbB7XBVkxCzANBgkqhkiG9w0BAQsFADBfMRIwEAYDVQQDDAlsb2NhbGhvc3QxDjAMBgNVBAgMBVRFWEFTMQ8wDQYDVQQHDAZBVVNUSU4xDTALBgNVBAsMBFBJTkcxDDAKBgNVBAoMA0NEUjELMAkGA1UEBhMCVVMwHhcNMjMwNzE0MDI1NDUzWhcNMjQwNzEzMDI1NDUzWjBfMRIwEAYDVQQDDAlsb2NhbGhvc3QxDjAMBgNVBAgMBVRFWEFTMQ8wDQYDVQQHDAZBVVNUSU4xDTALBgNVBAsMBFBJTkcxDDAKBgNVBAoMA0NEUjELMAkGA1UEBhMCVVMwggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQC5yFrh9VR2wk9IjzMz+Ei80K453g1j1/Gv3EQ/SC9h7HZBI6aV9FaEYhGnaquRT5q87p8lzCphKNXVyeL6T/pDJOW70zXItkl8Ryoc0tIaknRQmj8+YA0Hr9GDdmYev2yrxSoVS7s5Bl8poasn3DljgnWT07vsQz+hw3NY4SPp7IFGP2PpGUBBIIvrOaDWpPGsXeznBxSFtis6Qo+JiEoaVql9b9/XyKZj65wOsVyZhFWeM1nCQITSP9OqOc9FSoDFYQ1AVogm4A2AzUrkMnT1SrN2dCuTmNbeVw7gOMqMrVf0CiTv9hI0cATbO5we1sPAlJxscSkJjsaI+sQfjiAnAgMBAAEwDQYJKoZIhvcNAQELBQADggEBACgwoH1qklPF1nI9+WbIJ4K12Dl9+U3ZMZa2lP4hAk1rMBHk9SHboOU1CHDQKT1Z6uxi0NI4JZHmP1qP8KPNEWTI8Q76ue4Q3aiA53EQguzGb3SEtyp36JGBq05Jor9erEebFftVl83NFvio72Fn0N2xvu8zCnlylf2hpz9x1i01Xnz5UNtZ2ppsf2zzT+4U6w3frH+pkp0RDPuoe9mnBF001AguP31hSBZyZzWcwQltuNELnSRCcgJl4kC2h3mAgaVtYalrFxLRa3tA2XF2BHRHmKgocedVhTq+81xrqj+WQuDmUe06DnrS3Ohmyj3jhsCCluznAolmrBhT/SaDuGg="
```
henryrecker-pingidentity, line 104: Is this just a locally generated cert?

internal/acctest/config/sp/idpconnection/sp_idp_connection_resource_gen_test.go
```diff
@@ -0,0 +1,548 @@
...
+}
+
+func TestAccSpIdpConnection_MinimalMaximal(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() { acctest.ConfigurationPreCheck(t) },
+		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
+			"pingfederate": providerserver.NewProtocol6WithError(provider.NewTestProvider()),
+		},
+		CheckDestroy: spIdpConnection_CheckDestroy,
+		Steps: []resource.TestStep{
```
henryrecker-pingidentity, line 50: Do you think it would be worth adding more steps to cover the various "types" of idp_connection individually?

internal/resource/config/sp/idpconnection/sp_idp_connection_resource.go
wesleymccollam: No default value on this one.

internal/resource/config/sp/idpconnection/sp_idp_connection_resource.go
```diff
@@ -0,0 +1,4529 @@
...
+												},
+											},
+										},
+									},
+								},
+								"attribute_contract_fulfillment": attributecontractfulfillment.ToSchema(true, false, false),
+								"attribute_sources":              attributesources.ToSchema(0, false),
+								"issuance_criteria": schema.SingleNestedAttribute{
+									Description: "The issuance criteria that this transaction must meet before the corresponding attribute contract is fulfilled.",
+									// Computed:    true,
```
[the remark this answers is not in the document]
wesleymccollam: This one has no default value - the common package has a default set. Should we add a flag?

internal/resource/config/sp/idpconnection/sp_idp_connection_resource.go
henryrecker-pingidentity: I'm surprised they wouldn't be able to use the same default, since it's just defaulting conditional_criteria to an empty set.

internal/json/conversions.go
```diff
@@ -18,7 +18,8 @@ func FromValue(value attr.Value, skipNullOrUnknownAttrs bool) string {
 	if ok {
 		jsonString.WriteRune('"')
 		// Ensure any escaped quotes in the string are handled so that the resulting json includes a backslash
```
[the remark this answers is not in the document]
henryrecker-pingidentity: The new comment isn't quite right, the newlines aren't being removed they are just being escaped

internal/resource/config/sp/idpconnection/sp_idp_connection_resource.go
henryrecker-pingidentity: This line is still in place

henryrecker-pingidentity: Still need this

internal/resource/common/connectioncert/connection_cert_schema.go
```diff
@@ -0,0 +1,178 @@
...
+						stringvalidator.OneOf("HSM", "LOCAL"),
+					},
+				},
+			},
+			PlanModifiers: []planmodifier.Object{
+				planmodifiers.ValidateX509FileData(),
+			},
+		},
+		"active_verification_cert": schema.BoolAttribute{
+			Description: "Indicates if this is the active verification certificate. Default is false.",
```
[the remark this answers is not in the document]
henryrecker-pingidentity: Still have defaults without backticks

henryrecker-pingidentity:
I'd remove lines like "To update the password, specify the plaintext value in this field." in the schema descriptions, because it's basically a given for how terraform works.

henryrecker-pingidentity:
I'd check that there are no inconsistent result errors when you provide an empty `attribute_contract`, since PF will generally return always an empty list for extended_attributes if none is provided in the request.

henryrecker-pingidentity:
Reference to `eventTrigger` should use terraform formatting for the attribute name and the value in description for "stored_procedure"

henryrecker-pingidentity:
Missing enum options in description for jit_repository_attribute_mapping.source.type

henryrecker-pingidentity:
`front_channel_logout_uri` claims to be read-only but the schema marks it as optional

henryrecker-pingidentity:
One of the `source.type` attributes is missing the optional enum values in its description

internal/acctest/config/sp/idpconnection/sp_idp_connection_resource_gen_test.go
wesleymccollam: yes

henryrecker-pingidentity (approved):

internal/resource/config/sp/idpconnection/sp_idp_connection_resource.go
```diff
@@ -0,0 +1,4823 @@
...
+													MarkdownDescription: "A key that is meant to reference a source from which an attribute can be retrieved. This model is usually paired with a value which, depending on the SourceType, can be a hardcoded value or a reference to an attribute name specific to that SourceType. Not all values are applicable - a validation error will be returned for incorrect values.<br>For each SourceType, the value should be:<br>ACCOUNT_LINK - If account linking was enabled for the browser SSO, the value must be 'Local User ID', unless it has been overridden in PingFederate's server configuration.<br>ADAPTER - The value is one of the attributes of the IdP Adapter.<br>ASSERTION - The value is one of the attributes coming from the SAML assertion.<br>AUTHENTICATION_POLICY_CONTRACT - The value is one of the attributes coming from an authentication policy contract.<br>LOCAL_IDENTITY_PROFILE - The value is one of the fields coming from a local ide
```
henryrecker-pingidentity, line 4781: This needs to use `config.AuthContext`. Same for Read and Create

patrickcping (commented):
PDI-1944

PR #382, opened 2024-09-11
add deprecation message

wesleymccollam:


internal/resource/config/serversettings/systemkeys/server_settings_system_keys_resource.go
```diff
@@ -52,7 +52,8 @@ type serverSettingsSystemKeysResource struct {
 // GetSchema defines the schema for the resource.
 func (r *serverSettingsSystemKeysResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
 	schema := schema.Schema{
-		Description: "Manages the system encryption keys.",
+		Description:        "Manages the system encryption keys.",
+		DeprecationMessage: "This resource is deprecated and will be removed in a future release.",
```
henryrecker-pingidentity, line 56: We should describe which resource to use instead in this message, as well as updating the docs and the changelog to mention the deprecation.

henryrecker-pingidentity (approved):

PR #407, opened 2024-09-24
Upgrade guide v0 to v1

patrickcping:


docs/guides/version-1-upgrade.md
```diff
@@ -0,0 +1,369 @@
...
+
+### Removal of previously deprecated resources/data sources
+
+Removal of deprecated resources / data sources are expected on each major release going forward.  Ping maintains a deprecation and release strategy according to [Terraform provider creation best practices](https://developer.hashicorp.com/terraform/plugin/best-practices/versioning) and support of PingFederate product versions aligns with the [Ping Identity End of Life Policy](https://www.pingidentity.com/en/legal/end-of-life-policy.html).
+
+### Removal of support for previous PingFederate versions
+
+The support of PingFederate product versions aligns with the [Ping Identity End of Life Policy](https://www.pingidentity.com/en/legal/end-of-life-policy.html).  In this major release, supported PingFederate versions remain unchanged, versions `11.2` through `12.1` of PingFederate are supported.
```
patrickcping: @henryrecker-pingidentity just checking here that 11.2 is not EOL and that 12.1 is the latest we'll support at the point of `v1.0` release

docs/guides/version-1-upgrade.md
```diff
@@ -0,0 +1,369 @@
...
+
+### Removal of previously deprecated resources/data sources
+
+Removal of deprecated resources / data sources are expected on each major release going forward.  Ping maintains a deprecation and release strategy according to [Terraform provider creation best practices](https://developer.hashicorp.com/terraform/plugin/best-practices/versioning) and support of PingFederate product versions aligns with the [Ping Identity End of Life Policy](https://www.pingidentity.com/en/legal/end-of-life-policy.html).
+
+### Removal of support for previous PingFederate versions
+
+The support of PingFederate product versions aligns with the [Ping Identity End of Life Policy](https://www.pingidentity.com/en/legal/end-of-life-policy.html).  In this major release, supported PingFederate versions remain unchanged, versions `11.2` through `12.1` of PingFederate (supported in the `v0.*` release) are still supported.
```
patrickcping, line 33: @henryrecker-pingidentity just checking here that `11.2` is not EOL and that `12.1` is the latest we'll support at the point of `v1.0` release

patrickcping:
@henryrecker-pingidentity Need a review for whether this guide contains all the breaking changes we intend to make:
- Removal of fields (both deprecated and not deprecated)
- Removal of resources, data-sources (both deprecated and not deprecated)
- Changes to any data types leading to customers having to change their HCL (these need state upgraders)
- Renaming fields leading to customers having to change their HCL (these need state upgraders)

Also need to check whether I've added something that is deprecated that we intend to keep in the provider because of backward compatibility with previous PF versions

docs/guides/version-1-upgrade.md
henryrecker-pingidentity: Correct, those are the versions we'll be supporting

henryrecker-pingidentity:
> @henryrecker-pingidentity Need a review for whether this guide contains all the breaking changes we intend to make:
> 
> * Removal of fields (both deprecated and not deprecated)
> * Removal of resources, data-sources (both deprecated and not deprecated)
> * Changes to any data types leading to customers having to change their HCL (these need state upgraders)
> * Renaming fields leading to customers having to change their HCL (these need state upgraders)
> 
> Also need to check whether I've added something that is deprecated that we intend to keep in the provider because of backward compatibility with previous PF versions

A few attributes have slightly different underscore naming in idp_sp_connection vs sp_idp_connection, so I plan to change those for 1.0, so that will need to be added here when it is done.

I'm not aware of anything marked as deprecated that we intend to keep in 1.0, or of any changes to data types.

henryrecker-pingidentity (approved):

PR #411, opened 2024-09-26
corrected HCL examples following #408

patrickcping:


henryrecker-pingidentity (approved):
Good catch, thanks Patrick

PR #432, opened 2024-12-06
idp_sp_connection test improvements and minor fixes

henryrecker-pingidentity:
- Add a test for each type of sp connection
- Fix many missing defaults
- Fix broken validation for a couple booleans
- Fix some false booleans being left out of state


wesleymccollam (approved):

patrickcping (changes requested):
CDI-38

PR #493, opened 2025-04-18
Update 5 older acceptance tests

henryrecker-pingidentity:
- certificate_ca
- data_store (jdbc, custom, ldap, p1 gateway)


henryrecker-pingidentity:
Added back in the pingone connection to data.json because building it in-test is prone to conflicts

PR #527, opened 2025-07-14
Fix unexpected empty plans for sso_application_endpoint in pingfederate_sp_idp_connection

henryrecker-pingidentity:
This field had empty plans generated due to it not being given a value with certain IdP connection configurations. Now unknown will be treated as null by ModifyPlan when there is no expected change in this value.


henryrecker-pingidentity:
Closes #525

PR #530, opened 2025-07-23
Allow parallel test runs in scheduled tests

henryrecker-pingidentity:
This will speed up test times when covering multiple versions


henryrecker-pingidentity:
Successful run, much faster https://github.com/pingidentity/terraform-provider-pingfederate/actions/runs/16474267701

PR #546, opened 2025-08-21
Change attribute_sources to an ordered list, rather than an unordered set

henryrecker-pingidentity:
This change aligns with UI behavior for attribute sources


henryrecker-pingidentity:
Closing to be revisited later. CDI-591

PR #594, opened 2026-04-21
fix(#590): handle server-injected binaryAttributeSettings on LDAP attribute sources

### LINKED ISSUE #590: pingfederate_openid_connect_policy: attribute_mapping.attribute_sources with ldap_attribute_source produces "does not correlate with any element in actual"
### Community Note

* Please vote on this issue by adding a 👍 reaction to help the community and maintainers prioritize this request

### Affected Resource(s)

- `pingfederate_oauth_access_token_mapping`
- `pingfederate_openid_connect_policy`
- Likely all resources using `attribute_sources` with `ldap_attribute_source`

### terraform-provider-pingfederate Version

v1.7.1 (latest)

### PingFederate Version

Tested on both PF 11.3.7.1 and PF 13.0.1.0 — same result.

### OpenTofu/Terraform Version

OpenTofu v1.9.x

### Expected Behavior

When creating a resource with `attribute_sources` containing an `ldap_attribute_source`, the resource should be created and the state should reconcile cleanly. A subsequent `plan` should show no changes.

### Actual Behavior

The resource IS created successfully in PingFederate (confirmed via admin API), but the provider fails to reconcile the response with the planned state. The error is:

```
Error: Provider produced inconsistent result after apply

.attribute_sources: planned set element
cty.ObjectVal(map[string]cty.Value{
  "custom_attribute_source": cty.NullVal(...),
  "jdbc_attribute_source": cty.NullVal(...),
  "ldap_attribute_source": cty.ObjectVal(...)
}) does not correlate with any element in actual.

This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

This appears to be a terraform-plugin-framework issue where `attribute_sources` is defined as a Set with optional nested objects (`ldap_attribute_source`, `jdbc_attribute_source`, `custom_attribute_source`). The Set element correlation fails because the planned value (with null optional attributes) doesn't match the actual value returned by the API.

### Steps to Reproduce

1. Create a `pingfederate_oauth_access_token_mapping` with an `attribute_sources` block containing an `ldap_attribute_source`
2. Run `tofu apply` (or `terraform apply`)
3. The resource is created in PF but the provider errors

Minimal reproduction for `pingfederate_oauth_access_token_mapping`:

```hcl
resource "pingfederate_oauth_access_token_mapping" "test" {
  context = {
    type = "AUTHENTICATION_POLICY_CONTRACT"
    context_ref = { id = "your-apc-id" }
  }
  access_token_manager_ref = { id = "your-atm-id" }

  attribute_sources = [
    {
      ldap_attribute_source = {
        data_store_ref         = { id = "your-ldap-data-store-id" }
        id                     = "AD"
        description            = "AD"
        type                   = "LDAP"
        base_dn                = "dc=example,dc=com"
        search_scope           = "SUBTREE"
        search_filter          = "(sAMAccountName=${apc.sAMAccountName})"
        search_attributes      = ["Subject DN", "memberOf", "mail"]
        member_of_nested_group = false
      }
    }
  ]

  attribute_contract_fulfillment = {
    "groups" = {
      source = { type = "EXPRESSION" }
      value  = "#this.get(\"ds.AD.memberOf\")"
    }
  }
}
```

The same bug also affects `pingfederate_openid_connect_policy` when `attribute_mapping.attribute_sources` contains an `ldap_attribute_source`.

### Impact

This bug prevents any Terraform/OpenTofu-managed resource from using LDAP attribute sources in attribute mappings. The resource is created in PF but Terraform state is corrupted, causing errors on every subsequent plan/apply. This blocks use cases like OGNL-based group filtering with LDAP lookups in OIDC policies and access token mappings.

### Workaround

Create resources with LDAP attribute sources manually via the PF admin UI or API, and reference them by ID in Terraform without managing the `attribute_sources` block. For access token mappings, use `AUTHENTICATION_POLICY_CONTRACT` pass-through instead of LDAP sources (loses OGNL filtering capability).

### References

- Related upstream framework issue pattern: terraform-plugin-framework Set type with optional nested objects
- The root cause appears to be that Set element correlation uses structural equality, but the planned value has `null` for unused optional nested objects while the actual value may have a different representation

cdhesse:
### Summary

  Fixes #590 — `attribute_sources` with `ldap_attribute_source` produces "does not correlate with any element in actual"
  error on `pingfederate_openid_connect_policy` and `pingfederate_oauth_access_token_mapping`.

  ### Root Cause

  The PingFederate API injects `binaryAttributeSettings` into the POST/PUT response for LDAP attribute sources even when
  the user didn't specify it. Since `attribute_sources` is a Set, the terraform-plugin-framework uses structural
  equality for element correlation. The planned Set element has `null` for `binary_attribute_settings` while the actual
  value from the API response has an empty/populated map, causing correlation to fail.

  ### Changes

  **`attribute_sources_schema.go`**
  - Mark `binary_attribute_settings` as `Computed` (was `Optional` only) so the framework accepts server-injected values
  - Add `UseStateForUnknown()` plan modifier to prevent perpetual diffs on subsequent plans
  - Remove `SizeAtLeast(1)` validator which would reject an empty computed map

  **`openid_connect_policy_resource.go`**
  - Use a follow-up GET after Create/Update instead of the write response to populate state, ensuring the state matches
  the canonical API representation

  **`oauth_access_token_mapping_resource.go`**
  - Same read-after-write pattern as above (this resource is also called out in #590)

  ### Notes

  I noticed the `pingfederate_oauth_ciba_server_policy_request_policy-attribute_sources-plannedSetElementBugFix` branch
  is working on related `attribute_sources` Set correlation issues. This PR is a minimal subset targeting the
  LDAP-specific regression reported in #590. Happy to coordinate if there's overlap.


henryrecker-pingidentity:
Internal ticket CDI-893 will track our fix for this, this runs into issues with defaults/computeds in sets in the plugin framework.

PR #609, opened 2026-06-15
(fix)resource/redirect_validation: boolean attributes inside white_list and uri_allow_list sets change on each apply

wesleymccollam:
Change white_list and uri_allow_list from Set to List in redirect_validation resource

Fixes CDI-1076: boolean attributes inside white_list and uri_allow_list set entries were flipping between plans due to a known terraform-plugin-framework bug where Default values on nested set attributes cause element hashes to change each cycle. Converting to ordered lists resolves the flipping and restores Default values on nested attributes for proper zero-value handling. Adds index-based test assertions to cover list ordering and the CDI-1076 regression.

### DIFF
--- a/.changelog/pr-609.txt
+++ b/.changelog/pr-609.txt
@@ -1,0 +1,7 @@
+```release-note:breaking-change
+`resource/pingfederate_redirect_validation`: Updated the `redirect_validation_local_settings.white_list` and `redirect_validation_local_settings.uri_allow_list` fields to be Lists instead of Sets. This change helps prevent errors due to known terraform issues with nested sets. This change will not require any change to existing configuration, but existing state will need to be re-imported. These fields are now ordered instead of unordered, so future changes in ordering in configuration will result in planned changes with terraform.
+```
+
+```release-note:bug
+`resource/pingfederate_redirect_validation`: Fixed an issue where boolean attributes inside `white_list` and `uri_allow_list` entries would appear to flip between plans after each apply. Related to a known terraform-plugin-framework bug with `Default` values in nested set attributes: [#867](https://github.com/hashicorp/terraform-plugin-framework/issues/867).
+```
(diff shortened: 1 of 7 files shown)


.changelog/pr-609.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:breaking-change
+`resource/pingfederate_redirect_validation`: Updated the `redirect_validation_local_settings.white_list` and `redirect_validation_local_settings.uri_allow_list` fields to be Lists instead of Sets. This change helps prevent errors due to known terraform issues with nested sets. This change will not require any change to existing configuration, but existing state will need to be re-imported. These fields are now ordered instead of unordered, so future changes in ordering in configuration will result in planned changes with terraform.
```
henryrecker-pingidentity: "but existing state will need to be re-imported" should not be necessary, unless you ran into issues with this while testing? See https://github.com/pingidentity/terraform-provider-pingfederate/blob/36ffd57250eea2e53814769cbda06cfd49d68445/.changelog/archive/v1.8.0/pr-602.txt#L2

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# Unreleased
```
henryrecker-pingidentity: We don't maintain this file any more - the process is now to use the `release-notes` directory

internal/acctest/config/redirectvalidation/redirect_validation_resource_gen_test.go
```diff
@@ -28,8 +28,9 @@ func TestAccRedirectValidation_MinimalMaximal(t *testing.T) {
 				Check:  redirectValidation_CheckComputedValuesMinimal(),
 			},
 			{
-				// Update to a complete model. No computed values to check.
+				// Update to a complete model; verify list ordering by index
 				Config: redirectValidation_CompleteHCL(),
+				Check:  redirectValidation_CheckComputedValuesComplete(),
```
henryrecker-pingidentity: This check isn't really necessary as the previous comment said, because there are no computed values. https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.16.0/helper/resource#TestCheckResourceAttr

henryrecker-pingidentity (changes requested):

henryrecker-pingidentity (approved):

PR #630, opened 2026-08-17
Bump to newer PF patches for test images

henryrecker-pingidentity:


henryrecker-pingidentity:
Final working pipeline: https://github.com/pingidentity/terraform-provider-pingfederate/actions/runs/32060472099

internal/acctest/config/captchaproviders/captcha_provider_resource_gen_test.go
```diff
@@ -58,8 +58,19 @@ func TestAccCaptchaProvider_RemovalDrift(t *testing.T) {
 }
 
 func TestAccCaptchaProvider_MinimalMaximal(t *testing.T) {
+	connId := os.Getenv("PF_TF_P1_CONNECTION_ID")
```
henryrecker-pingidentity, line 61: These were previously only checked in the RemovalDrift test above, so this test always failed when run solo

PR #634, opened 2026-09-09
fix(oauth_access_token_manager): make sequence_number configurable

wesleymccollam:
Fixes CDI-1452

## Summary

`sequence_number` on `pingfederate_oauth_access_token_manager` was defined as `Computed`-only, so Terraform could not set it — even though the PingFederate Admin API accepts `sequenceNumber` on both POST and PUT. Sequence numbers identify which token manager issued a token (embedded in the `pi.atm` JWT header claim), so customers managing token managers declaratively across Dev/Staging/Prod need to pin them to avoid drift and token-validation failures.

## Changes

- `internal/resource/config/oauth/accesstokenmanager/oauth_access_token_manager_resource_gen.go`: `sequence_number` is now `Optional` in addition to `Computed`. No validator is added — the API documents no constraint on the field, and the provider's convention is to add plan-time validators only where the API documents a bound (all 38 existing `int64validator` usages map to a documented bound; the ~180 other int attributes are unvalidated). Values outside the server's accepted range are rejected by PingFederate with a 422. The existing `UseNonNullStateForUnknown` plan modifier is unchanged, so removing the attribute from configuration retains the previously assigned value instead of diffing.
- Acceptance test `TestAccOauthAccessTokenManager_SequenceNumber`: create pinned (`123456`) → update (`654321`) → remove from config (with a name change so the update path runs) and verify the value persists → full config reshape still persists → import-verify.
- Regenerated `docs/resources/oauth_access_token_manager.md` (`sequence_number` moves Read-Only → Optional). Data-source doc intentionally unchanged.

## API evidence (live PF 13.1)

Captured in `scratch/cdi-1452-sequence-number/` (untracked):

| Call | Response `sequenceNumber` |
|---|---|
| POST with `424242` | `424242` (echoed) |
| PUT with `424243` | `424243` (echoed) |
| PUT omitting the field | prior value `424243` preserved |
| POST omitting the field | server-assigned random (`1429725`) |

The echo confirms Create/Update can set state from the write-response body — no GET-after-write needed. The OpenAPI spec declares the field `integer/int32` with no `readOnly` annotation and no constraints.

## Validation

- `go build ./...`, `go vet`, `golangci-lint`, `tfproviderlint`, `tflint`, `verifycontent` all pass
- Full ATM acceptance suite passes against PF 13.1 (`RemovalDrift`, `MinimalMaximalInternallyManaged`, `SequenceNumber`, `MinimalMaximalJsonWeb`); `TestAccOauthAccessTokenManagerSettings_MinimalMaximal` passes when run in isolation (pre-existing ordering sensitivity, unrelated to this change)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changelog/pr-634.txt
+++ b/.changelog/pr-634.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+`resource/pingfederate_oauth_access_token_manager`: Fixed an issue where `sequence_number` could not be configured. The attribute is now Optional in addition to Computed, allowing the sequence number (embedded in the `pi.atm` claim of issued access tokens) to be pinned to a specific value. If not configured, a server-assigned value is used, and the previously-assigned value is retained if the attribute is later removed from configuration.
+```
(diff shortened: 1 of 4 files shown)


henryrecker-pingidentity (commented):
The changes seem good to me - have you tested this on any versions prior to 13.1?

wesleymccollam:
> The changes seem good to me - have you tested this on any versions prior to 13.1?

verified 12.2.6, 12.3.6, 13.0.4 as well

henryrecker-pingidentity (approved):

PR #635, opened 2026-09-10
CDI-1474: Add validators to prevent inconsistent result errors

henryrecker-pingidentity:
- Add new attribute contract fulfillment validator logic to raise a validator error on expressions with trailing newlines, since the PF admin API trims them automatically
- Add new validator for expression_criteria requiring length of at least 1, since PF treats an empty expression_criteria as equivalent to null
- Add these validators across resources

Passing test pipeline: https://github.com/pingidentity/terraform-provider-pingfederate/actions/runs/34503911591

### DIFF
--- a/.changelog/pr-TBD.txt
+++ b/.changelog/pr-TBD.txt
@@ -1,3 +1,0 @@
-```release-note:bug
-Added new validation to prevent two "Provider produced inconsistent result after apply" errors: (1) `issuance_criteria.expression_criteria` now must contain at least one element if specified, as the API treats an empty `expression_criteria` as equivalent to `null`. (2) `attribute_contract_fulfillment` `value` must no longer end in a trailing newline when `source.type` is `EXPRESSION` — the PingFederate admin API trims trailing newlines from OGNL expressions, which causes inconsistent result errors for values written using Terraform heredoc (`<<-EOT`) syntax.
-```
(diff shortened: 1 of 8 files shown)


.changelog/pr-TBD.txt
jonathanfuller-pingidentity, line 1: pr-TBD.txt

.changelog/pr-TBD.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+Added new validation to prevent two "Provider produced inconsistent result after apply" errors: (1) `issuance_criteria.expression_criteria` now must contain at least one element if specified, as the API treats an empty `expression_criteria` as equivalent to `null`. (2) `attribute_contract_fulfillment` `value` must no longer end in a trailing newline when `source.type` is `EXPRESSION` — the PingFederate admin API trims trailing newlines from OGNL expressions, which causes inconsistent result errors for values written using Terraform heredoc (`<<-EOT`) syntax.
```
jonathanfuller-pingidentity: This is pretty long for a release note. Could any of the detail be off-loaded to docs?

jonathanfuller-pingidentity (approved):
Just changelog.

.changelog/pr-TBD.txt
henryrecker-pingidentity: Shortened this

